PR #48, opened 2018-08-14
Selective fields for NGrams

kashmawy:
1. Rename Sequence to NGram
2. Update Reader interface to take in schema_fields either the current form (iterable of Unischema Fields) or NGram
3. Update NGram to take a dictionary of fields, where the keys are consecutive integers and the values are an array of Unischema fields to return at each timestep AS OPPOSED TO returning all the fields at all timesteps.


rgruener:
closing due to new PR https://github.com/uber/petastorm/pull/66

PR #50, opened 2018-08-15
Pytorch example with DataLoader adapter, using MNIST data

forbearer:
This code includes an MNIST dataset generator, a pytorch training example that uses the resulting dataset, and a simple `README.md`.

As can be seen from the `main.py`, there are few _limitations_ that come to light which could help us improve petastorm:
* Batch shuffling
* Support for custom transforms
* Total data size (or some semblance of it?)

Running `pytorch/examples/mnist/main.py` (in a Docker container) with the default 10 epoch yielded the following outcome (I just show the test output for the middle 8 epochs):
```
...
Train Epoch: 1 [59520/60000 (99%)]	Loss: 0.505042

Test set: Average loss: 0.2056, Accuracy: 9395/10000 (94%)

...
Test set: Average loss: 0.1337, Accuracy: 9596/10000 (96%)
Test set: Average loss: 0.1033, Accuracy: 9684/10000 (97%)
Test set: Average loss: 0.0919, Accuracy: 9710/10000 (97%)
Test set: Average loss: 0.0760, Accuracy: 9770/10000 (98%)
Test set: Average loss: 0.0689, Accuracy: 9797/10000 (98%)
Test set: Average loss: 0.0623, Accuracy: 9803/10000 (98%)
Test set: Average loss: 0.0632, Accuracy: 9791/10000 (98%)
Test set: Average loss: 0.0541, Accuracy: 9818/10000 (98%)

...
Train Epoch: 10 [59520/60000 (99%)]	Loss: 0.040862

Test set: Average loss: 0.0505, Accuracy: 9845/10000 (98%)

real	3m3.021s
user	20m4.680s
sys	0m22.228s
```
With the petastormed variant, the training accuracy looks on-par, with somewhat better runtime. I'll show just the test output:
```
Test set: Average loss: 0.2035, Accuracy: 9385/10000 (94%)
Test set: Average loss: 0.1326, Accuracy: 9591/10000 (96%)
Test set: Average loss: 0.1040, Accuracy: 9675/10000 (97%)
Test set: Average loss: 0.0887, Accuracy: 9705/10000 (97%)
Test set: Average loss: 0.0761, Accuracy: 9752/10000 (98%)
Test set: Average loss: 0.0715, Accuracy: 9774/10000 (98%)
Test set: Average loss: 0.0627, Accuracy: 9797/10000 (98%)
Test set: Average loss: 0.0606, Accuracy: 9810/10000 (98%)
Test set: Average loss: 0.0582, Accuracy: 9824/10000 (98%)
Test set: Average loss: 0.0548, Accuracy: 9828/10000 (98%)

real	2m35.852s
user	2m33.508s
sys	0m6.576s
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-dist: trusty
+dist: xenial
 language: python
 cache: pip
 python:
@@ -31,7 +31,10 @@ install:
 
 script:
   - pylint --rcfile=.pylintrc petastorm examples -f parseable -r n
-  - pytest -v --cov=./
+  # Completely ignore two pytorch tests to prevent static-TLS-caused torch-import seg-fault
+  - pytest -v --ignore=petastorm/tests/test_pytorch_utils.py --ignore=examples/mnist/tests/test_generate_mnist_dataset.py --cov=./
+  # Run the pytorch tests separately, but accumulate code coverage data
+  - pytest -v examples/mnist/tests/test_generate_mnist_dataset.py petastorm/tests/test_pytorch_utils.py --cov=./ --cov-append
 
   # Verify caching of synthetic dataset is working (at least not failing)
   - pytest --cache-synthetic-dataset --log-cli-level info petastorm/tests/test_end_to_end.py::test_reading_subset_of_columns
--- a/README.md
+++ b/README.md
@@ -139,12 +139,10 @@ def _transform_row(mnist_row):
     ])
     return (transform(mnist_row.image), mnist_row.digit)
 
-with DataLoader(Reader('file:///localpath/mnist/train', reader_pool=ThreadPool(1),
-                       shuffle_options=ShuffleOptions(), num_epochs=10),
+with DataLoader(Reader('file:///localpath/mnist/train', num_epochs=10),
                 batch_size=64, transform=_transform_row) as train_loader:
     train(model, device, train_loader, 10, optimizer, 1)
-with DataLoader(Reader('file:///localpath/mnist/test', reader_pool=ThreadPool(1),
-                       shuffle_options=ShuffleOptions(), num_epochs=10),
+with DataLoader(Reader('file:///localpath/mnist/test', num_epochs=10),
                 batch_size=1000, transform=_transform_row) as test_loader:
     test(model, device, test_loader)
 ```
--- a/examples/mnist/main.py
+++ b/examples/mnist/main.py
@@ -27,8 +27,7 @@ import torch.optim as optim
 from torchvision import transforms
 
 from petastorm.pytorch import DataLoader
-from petastorm.reader import Reader, ShuffleOptions
-from petastorm.workers_pool.thread_pool import ThreadPool
+from petastorm.reader import Reader
 
 
 class Net(nn.Module):
@@ -144,12 +143,10 @@ def main():
 
     # Instantiate each petastorm Reader with a single thread, shuffle enabled, and appropriate epoch setting
     for epoch in range(1, loop_epochs + 1):
-        with DataLoader(Reader('{}/train'.format(args.dataset_url), reader_pool=ThreadPool(1),
-                               shuffle_options=ShuffleOptions(), num_epochs=reader_epochs),
+        with DataLoader(Reader('{}/train'.format(args.dataset_url), num_epochs=reader_epochs),
                         batch_size=args.batch_size, transform=_transform_row) as train_loader:
             train(model, device, train_loader, args.log_interval, optimizer, epoch)
-        with DataLoader(Reader('{}/test'.format(args.dataset_url), reader_pool=ThreadPool(1),
-                               shuffle_options=ShuffleOptions(), num_epochs=reader_epochs),
+        with DataLoader(Reader('{}/test'.format(args.dataset_url), num_epochs=reader_epochs),
                         batch_size=args.test_batch_size, transform=_transform_row) as test_loader:
             test(model, device, test_loader)
 
--- a/examples/mnist/tests/test_generate_mnist_dataset.py
+++ b/examples/mnist/tests/test_generate_mnist_dataset.py
@@ -23,7 +23,7 @@ import numpy as np
 from examples.mnist.generate_petastorm_mnist import download_mnist_data, \
     mnist_data_to_petastorm_dataset
 from petastorm.reader import Reader, ShuffleOptions
-from petastorm.workers_pool.thread_pool import ThreadPool
+from petastorm.workers_pool.dummy_pool import DummyPool
 
 logging.basicConfig(level=logging.INFO)
 
@@ -113,31 +113,25 @@ def test_mnist_download(tmpdir):
     assert o[len(o)-1][1] == 6
 
 
-def test_generate_mnist_dataset(small_mock_mnist_data, tmpdir_factory):
-    tmpdir = generate_mnist_dataset(small_mock_mnist_data, tmpdir_factory)
-
-    train_path = os.path.join(tmpdir, 'train')
+def test_generate_mnist_dataset(generate_mnist_dataset):
+    train_path = os.path.join(generate_mnist_dataset, 'train')
     assert os.path.exists(train_path)
     assert os.path.exists(os.path.join(train_path, '_common_metadata'))
     assert os.path.exists(os.path.join(train_path, '_metadata'))
 
-    test_path = os.path.join(tmpdir, 'test')
+    test_path = os.path.join(generate_mnist_dataset, 'test')
     assert os.path.exists(test_path)
     assert os.path.exists(os.path.join(test_path, '_common_metadata'))
     assert os.path.exists(os.path.join(test_path, '_metadata'))
 
 
-@pytest.mark.skipif(sys.version_info < (3, 0),
-                    reason='unstable with chance of seg fault under Python 2.x, issue #52')
 def test_read_mnist_dataset(generate_mnist_dataset):
     # Verify both datasets via a reader
     for dset in SMALL_MOCK_IMAGE_COUNT.keys():
-        with Reader('file://{}/{}'.format(generate_mnist_dataset, dset), reader_pool=ThreadPool(1)) as reader:
+        with Reader('file://{}/{}'.format(generate_mnist_dataset, dset), reader_pool=DummyPool()) as reader:
             assert len(reader) == SMALL_MOCK_IMAGE_COUNT[dset]
 
 
-@pytest.mark.skipif(sys.version_info < (3, 0),
-                    reason='unstable with chance of seg fault under Python 2.x, issue #52')
 def test_full_pytorch_example(large_mock_mnist_data, tmpdir):
     # First, generate mock dataset
     dataset_url = 'file://{}'.format(tmpdir)
@@ -153,11 +147,9 @@ def test_full_pytorch_example(large_mock_mnist_data, tmpdir):
     model = main.Net().to(device)
     optimizer = torch.optim.SGD(model.parameters(), lr=0.01, momentum=0.5)
 
-    with DataLoader(Reader('{}/train'.format(dataset_url), reader_pool=ThreadPool(1),
-                           shuffle_options=ShuffleOptions(), num_epochs=1),
+    with DataLoader(Reader('{}/train'.format(dataset_url), reader_pool=DummyPool(), num_epochs=1),
                     batch_size=32, transform=main._transform_row) as train_loader:
         main.train(model, device, train_loader, 10, optimizer, 1)
-    with DataLoader(Reader('{}/test'.format(dataset_url), reader_pool=ThreadPool(1),
-                           shuffle_options=ShuffleOptions(), num_epochs=1),
+    with DataLoader(Reader('{}/test'.format(dataset_url), reader_pool=DummyPool(), num_epochs=1),
                     batch_size=100, transform=main._transform_row) as test_loader:
         main.test(model, device, test_loader)
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -152,7 +152,7 @@ class Reader(object):
             else:
                 logger.warning('shuffle option is deprecated. Please use shuffle_options instead')
             shuffle_options = ShuffleOptions(shuffle)
-        self._normalize_shuffle_options(shuffle_options, dataset)
+        self._normalize_shuffle_options(shuffle_options, self.dataset)
         ventilator = self._create_ventilator(filtered_row_group_indexes, shuffle_options, num_epochs, worker_predicate)
 
         # 5. Start workers pool
--- a/doc/troubleshoot.md
+++ b/doc/troubleshoot.md
@@ -1,0 +1,41 @@
+
+## Trouble running the pytorch example
+
+If you see the following error while trying to run the pytorch example, you are in luck:
+```bash
+  File "/usr/local/lib/python2.7/dist-packages/torch/__init__.py", line 80, in <module>
+    from torch._C import *
+ImportError: dlopen: cannot load any more object with static TLS
+```
+
+This problem stems from a known defect in glibc `dlopen` logic that made conservative
+assumptions about static thread-local storage, specific with respect to surplus
+DTV slots, which no longer suffice for modern compute needs.
+
+Solutions, in increasing order of effort involved:
+1. Do `import torch` as early as possible ([ref](https://github.com/tensorflow/models/issues/523#issuecomment-272754029))
+2. Install `libgomp1` and `export LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libgomp.so.1`
+3. Build pytorch from source ([ref](https://github.com/pytorch/pytorch/issues/643))
+4. Patch glibc to increase surplus DTV slots from 14 to 32 or 64.
+
+For background, this issue was reported back in 2013 with pMatlab since 2012](https://stackoverflow.com/a/19468365).
+For additional references, find the [glibc bug report and fix](https://sourceware.org/bugzilla/show_bug.cgi?id=17620),
+and the accompanying Debian glibc bug report [#793689](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793689).
+According to [#793641](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793641),
+some variabnt of the static TLS fix was included in glibc-2.22.
+The OpenMP library libgomp.so.1 has had this fix in place since circa 2015.
+Ubuntu Xenial and above also contains this fix, but just updating operating system
+may not be sufficient if `torch` links against its own version of glibc that
+still uses static TLS.
+An `ldd` analysis (per [#793689 common #20](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793689#24))
+can reveal whether libraries like torch is __actually__ still using static TLS.
+
+### Building pytorch from Dockerfile and using it
+
+If you choose to build pytorch from source, you can do so using the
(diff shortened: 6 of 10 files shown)


forbearer:
@selitvin About DataLoader and num_epochs, you are right that it's not in the constructor.  I had `num_workers` in my memory and that got mixed up when I was talking to you.  My bad.  This does simplify things.  :-)

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -0,0 +1,86 @@
...
+
+    np.set_printoptions(linewidth=200)
+    reshaped = np.array(list(im[0].getdata()), dtype=np.uint8).reshape(MOCK_IMAGE_3DIM_SIZE)
+    print(reshaped)
+    assert reshaped.shape == MOCK_IMAGE_3DIM_SIZE
+
+
+# This test requires torch import, but that's causing dlopen to fail in Travis run. :(
+# So skipping it for now
+def _dont_test_mnist_download(tmpdir):
```
forbearer: https://github.com/uber/petastorm/blob/c59f0c8deefe2b1c6cbdf00b08260c22eadef165/examples/mnist/tests/test_generate_mnist_dataset.py#L72-L79

I've disabled this test for now due to issue dlopen'ing `torch` module in travis CI run.

If it's OK with you, I'd like to tackle that issue in a separate PR...as this one is getting quite long.

forbearer:
Aww yeah! I'd be happy to land this and we can improve on the test in a future PR?

petastorm/pytorch.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __iter__(self):
+        """
+        The Data Loader iterator stops the for-loop at the end of each epoch, but a subsequent for-loop
+        will instantiate a new Reader and yield more results, until the requested number of epoch has been
+        reached.  After that point, any subsequent call results in StopIteration, per iterator protocol.
+        """
+        batch = []
+        for row in self.reader:
+            batch.append(self.transform(row))
```
selitvin: This transform can be heavy. We need to think about pushing this down to a worker. Will be easier with the new reader implementation. Not for this PR.

petastorm/pytorch.py
```diff
@@ -0,0 +1,118 @@
...
+            if len(batch) == self.batch_size:
+                yield self.collate_fn(batch)
+                batch = []
+        if batch:
```
selitvin: pytorch's `DataLoader` has `drop_last` to define whether we want an incomplete batch. Should be simple to add the implementation here. Ok if you want as a separate CL.

.travis.yml
```diff
@@ -20,9 +20,11 @@ python:
   - '3.6'
 
 install:
+  # Upgrade pip to avoid weird failures, such as failing to install torchvision
+  - pip install --upgrade pip
```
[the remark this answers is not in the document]
forbearer: Sure. I think that makes sense. How about 18.0? :-)

petastorm/pytorch.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    def __init__(self, dataset_url, schema_fields=None, batch_size=1, shuffle_options=None,
+                 collate_fn=default_collate, transform=None,
+                 predicate=None, rowgroup_selector=None,
+                 num_workers=0, reader_pool=None, num_epochs=1, sequence=None,
+                 training_partition=None, num_training_partitions=None,
+                 read_timeout_s=None, cache=None):
+        """
+        Initializes a batch reader object, with a default collate and optional transform functions.
```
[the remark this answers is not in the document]
forbearer: oops. thx!

petastorm/pytorch.py
forbearer: :-) Agreed!

petastorm/pytorch.py
forbearer: I think that makes sense. Let me add that in a separate CL/PR.

.travis.yml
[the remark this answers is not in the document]
rgruener: Once were doing that it might make sense to install numpy 1.14.5 (though maybe this PR is large enough)

README.md
```diff
@@ -101,6 +107,43 @@ TODO: Reading a dataset from pytorch
...
+The minimalist example below assumes the definition of a `Net` class and `train` and `test` functions, shown in the mnist example:
+
+```python
+import torch
+from petastorm.pytorch import DataLoader
+
+torch.manual_seed(1)
+device = torch.device('cpu')
+model = Net().to(device)
+optimizer = torch.optim.SGD(model.parameters(), lr=0,01, momentum=0.5)
```
rgruener: is `lr=0,01` a typo?

examples/mnist/tests/test_generate_mnist_dataset.py
rgruener: You should skip this test using `@pytest.mark.skip` instead of changing the function name

petastorm/pytorch.py
```diff
@@ -0,0 +1,118 @@
...
+    This class iterates and returns items from the Reader in batches.
+
+    This loader can be used as a context manager, but it will terminate at the end of an epoch.
+    The context will invoke next_epoch() upon entry.
+
+    If not used as context manager, invoke the next_epoch() function at the start of each epoch, and
+    once more at the very end.
+    """
+
+    def __init__(self, dataset_url, schema_fields=None, batch_size=1, shuffle_options=None,
```
rgruener: This class copies a lot of logic and boilerplate from the Reader class. This will require keeping a lot of things in sync, e.g. arguments, etc. Is there a reason we dont pass in an instantiated reader object to the DataLoader or have it be a parent class?

rgruener: This would also make it more inline with what we do with tensorflow (have it be an adapter on top of the python reader instead of a different class)

petastorm/reader.py
```diff
@@ -284,6 +284,13 @@ def fetch(self, timeout=None):
         logger.warn(warning_message)
         return self._workers_pool.get_results(timeout=timeout)
 
+    def __len__(self):
```
rgruener: please add a unit test for this function

README.md
forbearer: Yeah, good eye! I just posted a fix. Thanks.

petastorm/pytorch.py
forbearer: Hmm, that's a good point.  I also didn't like reproducing reader params all over again.

Also, I did battle these two options out in my mind...didn't think to look at tf for an existing pattern!

OK, let me know reformulate this to pass in an instantiated reader.

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+
+@pytest.fixture(scope="session")
+def large_mock_mnist_data():
+    return _mock_mnist_data(LARGE_MOCK_IMAGE_COUNT)
+
+
+def test_image_to_numpy(small_mock_mnist_data):
+    """ Show output of image object reshaped as numpy array """
+    im = small_mock_mnist_data['train'][0]
+    print(im)
```
selitvin: Should not have raw prints in the test.

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+    print(im[1])
+    assert im[1] >= 0 and im[1] <= 9
```
selitvin: Nitpick: `0<=im[1]<=9`

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -0,0 +1,132 @@
...
+
+    np.set_printoptions(linewidth=200)
+    reshaped = np.array(list(im[0].getdata()), dtype=np.uint8).reshape(MOCK_IMAGE_3DIM_SIZE)
+    print(reshaped)
+    assert reshaped.shape == MOCK_IMAGE_3DIM_SIZE
+
+
+# This test requires torch import, but that's causing dlopen to fail in Travis run. :(
+# So skipping it for now
+def _dont_test_mnist_download(tmpdir):
```
selitvin: Consider using a standard unittest feature:
```
@unittest.skip('some description')
```

petastorm/pytorch.py
```diff
@@ -0,0 +1,118 @@
...
+                 collate_fn=default_collate, transform=None,
+                 predicate=None, rowgroup_selector=None,
+                 num_workers=0, reader_pool=None, num_epochs=1, sequence=None,
+                 training_partition=None, num_training_partitions=None,
+                 read_timeout_s=None, cache=None):
+        """
+        Initializes a data loader object, with a default collate and optional transform functions.
```
selitvin: A question: would it make sense to pass a reader instance as a parameter to the DatasetLoader instead of forwarding all arguments to it? Kinda same as pytorch's Loader takes an instance of a `Dataset`?

And on the same note, the `reader_pool` is not forwarded, which emphasizes the point :)

examples/mnist/tests/test_generate_mnist_dataset.py
forbearer: Ah, agreed.  This apparent slows down test run!

I'll use logging to allow getting output with `pytest -s`.

examples/mnist/tests/test_generate_mnist_dataset.py
forbearer: Yes. Pythonic!

petastorm/pytorch.py
forbearer: Agreed! @rgruener also commented on this exact same thing. Great minds! :-)

Posting an updated commit real soon that addresses all these review comments.  Thank you so much!

forbearer:
@selitvin @rgruener Thank you both for your detailed round-2 reviews.

I've addressed all the comments in the new commit, re-amended and pushed.  I'll be checking on the test to get them passing, given two issues I've encountered so far:
1) dlopen error when importing `torch`
2) seg fault when running either `test_generate` or `test_full_pytorch_example`

petastorm/reader.py
forbearer: Definitely. I added a new case to `test_end_to_end.py`.

Arguably, I should include a case to test raising `NotImplementedError`!

forbearer:
OK, the good news is that the CI passes with Python 3.x!

forbearer:
Lacking access to travis debug mode, I'm trying various `pytest` options to ferret out the issue.

One thought is that `pytest` execution order is presumably deterministic, and the collect is always failing on item 129, with 1 failing, which means the 130th test is bad (or maybe 131st).

These are test cases 129, 130, 131 when the full list of tests are enumerated:
```
petastorm/tests/test_predicates.py::PredicatesTest::test_inclusion PASSED                                                                                                                                       [ 65%]
petastorm/tests/test_predicates.py::PredicatesTest::test_list_inclusion PASSED                                                                                                                                  [ 66%]
petastorm/tests/test_predicates.py::PredicatesTest::test_negation PASSED                                                                                                                                        [ 66%]
```

forbearer:
Removing the various pytest runs as they weren't yielding additional information.

FYI, this generates a list of test functions, without running them:
```
pytest --collect-only | grep Function
```

I can't see why any of the first 3 test cases in `test_predicates.py` would cause seg fault...definitely no pytorch import either :-)

forbearer:
Filtering out `test_predicates.py` didn't help the pytest collect seg-fault.  That's not surprising.

I decided to try out filtering out `test_pytorch_utils.py`.

forbearer:
Well, ladies and gentlemen, `test_pytorch_utils.py` was the culprit of the segfault during pytest collect.

Next: figuring out why...

forbearer:
Wow, so based on this (https://github.com/pytorch/pytorch/issues/2083) and some experimentation in the travis docker container, I was able to resolve the 2.7 torch import issue (with dlopen running out of static TLS) by importing torch *way early* in conftest.py.

That's really awful, but solves all of the problem surrounding my unit tests needing to be skipped.

I'll post that change as part of this PR.  Let me know what you think.

forbearer:
Strangely, this test is reliably seg-faulting now, even when I do `pytest -v` locally:
```
petastorm/tests/test_end_to_end.py::test_simple_read[reader_factory1] /home/travis/.travis/job_stages: line 78:  2933 Segmentation fault      (core dumped) pytest -v --cov=./
```

forbearer:
The other lingering issue is captured in #52.  For this, I will skip the mnist generate and pytorch example test cases for Python 2.x.

forbearer:
Well, the experiment to upgrade to `xenial` and `bionic` did not yield positive result.

So I started wondering if the dlopen static TLS problem was actually incorporated into those two Ubuntu releases.

From this https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1673956, the bug was analyzed and bugfixed in https://sourceware.org/bugzilla/show_bug.cgi?id=17620

According to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793641, the above fix is in `glibc-2.22`.

According to ubuntu, xenial http://changelogs.ubuntu.com/changelogs/pool/main/g/glibc/glibc_2.23-0ubuntu10/changelog has `glibc-2.22`

Well, as I still encounter the same dlopen static TLS error in both xenial and bionic, why might that be?

forbearer:
LOL: https://travis-ci.com/uber/petastorm/jobs/141963848

I was so stoked that the build passed, but discovered upon closer examination that the problem persists: seg fault during collect.

At this point, here is what I have tried and what I know:
* So long as test_pytorch_utils.py is in the pytest path, collect fails, even if using `-k` to exclude
* Different versions of Ubuntu did not resolve the issue, tried: `trusty`, `xenial`, and `bionic`, despite the fact that `glibc-2.22` (which **is** in `xenial` onward) contains the static TLS fix (*by increasing DTV surplus slots from 14 to 32, or some slightly larger integer*)
* Taking Clearwater's advice, I tried installing libgomp1 and preceding pytest with `export LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libgomp.so.1"`, but that did not resolve either!
* Using these commands (adapted from https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=793689#24), I believe the culprint *may* be that the `torch-0.4.1` we are pip installing statically linked against a specific libgomp, and therefore override does not work!
  * If I forcibly symlink torch's libgomp, I get the GOMP_4.0 not found error
  * Build from source *should* work, according to https://github.com/pytorch/pytorch/issues/643#issuecomment-292744036
```
for lib in $(ldd $(find /usr -name _C.so |& grep -v denied) | grep "=>" | sed 's/.*=> \([/a-z0-9\._+-]*\) .*/\1/g' | sort); do echo "readelf -d -W $lib"; readelf -d -W "$lib"; done | egrep 'readelf|STATIC_TLS' | grep -B1 TLS

readelf -d -W /lib/x86_64-linux-gnu/libc.so.6
 0x000000000000001e (FLAGS)              STATIC_TLS
--
readelf -d -W /lib/x86_64-linux-gnu/libm.so.6
 0x000000000000001e (FLAGS)              STATIC_TLS
readelf -d -W /lib/x86_64-linux-gnu/libpthread.so.0
 0x000000000000001e (FLAGS)              STATIC_TLS
readelf -d -W /lib/x86_64-linux-gnu/librt.so.1
 0x000000000000001e (FLAGS)              STATIC_TLS
--
readelf -d -W /usr/local/lib/python2.7/dist-packages/torch/lib/libgomp-7bcb08ae.so.1
 0x000000000000001e (FLAGS)              STATIC_TLS
```

Below, I do LD_PRELOAD first to show that `_C.so` still pulls torch's libgomp:
```
$ export LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libgomp.so.1"; for lib in $(ldd $(find /usr -name _C.so |& grep -v denied) | grep "=>" | sed 's/.*=> \([/a-z0-9\._+-]*\) .*/\1/g' | sort); do echo "readelf -d -W $lib"; readelf -d -W "$lib"; done | egrep 'readelf|STATIC_TLS' | grep -B1 TLS; unset LD_PRELOAD
readelf: Error: 'libnvToolsExt-3965bdd0.so.1': No such file
readelf: Error: '=>': No such file
readelf: Error: '(0x00007ff8e5dd3000)': No such file
readelf -d -W /lib/x86_64-linux-gnu/libc.so.6
 0x000000000000001e (FLAGS)              STATIC_TLS
--
readelf -d -W /lib/x86_64-linux-gnu/libm.so.6
 0x000000000000001e (FLAGS)              STATIC_TLS
readelf -d -W /lib/x86_64-linux-gnu/libpthread.so.0
 0x000000000000001e (FLAGS)              STATIC_TLS
readelf -d -W /lib/x86_64-linux-gnu/librt.so.1
 0x000000000000001e (FLAGS)              STATIC_TLS
--
readelf -d -W /usr/local/lib/python2.7/dist-packages/torch/lib/libgomp-7bcb08ae.so.1
 0x000000000000001e (FLAGS)              STATIC_TLS
```

**Last ditch effort**: I'm going to try pip installing the [CPU torch](http://download.pytorch.org/whl/cpu/torch-0.4.1-cp27-cp27mu-linux_x86_64.whl), followed by building from source using 
[Dockerfile](https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile). If neither of these efforts work, then I'm going to split pytorch tests from the rest of the pytest in order to make progress.

This PR has sunk quite a bit of time.  :-\

examples/mnist/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM ubuntu:14.04.5
```
forbearer: This Dockerfile is not meant to be committed (unless we want to for some reason).

I'm just trying to track what I've done.

.travis.yml
```diff
@@ -12,24 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-dist: trusty
+dist: xenial
```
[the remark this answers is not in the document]
forbearer: Yes, this is a deliberate change.  I think we should proceed with xenial to be ahead of the game.

examples/mnist/Dockerfile
```diff
@@ -0,0 +1,42 @@
+FROM pytorch:latest
```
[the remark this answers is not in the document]
forbearer: Ah, yeah, that's actually a good point.

If there will be a tf example using MNIST, perhaps I should simply make a pytorch subdirectory. Then that would make things much clearer?

examples/mnist/main.py
```diff
@@ -0,0 +1,158 @@
...
+        loop_epochs = 1
+        reader_epochs = args.epochs
+    else:
+        # Test training accuracy after each epoch
+        loop_epochs = args.epochs
+        reader_epochs = 1
+
+    # Instantiate each petastorm Reader with a single thread, shuffle enabled, and appropriate epoch setting
+    for epoch in range(1, loop_epochs + 1):
+        with DataLoader(Reader('{}/train'.format(args.dataset_url), reader_pool=ThreadPool(1),
```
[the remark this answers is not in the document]
forbearer: Both arguments `reader_pool` and `shuffle_options` are specified in this code, or do you mean, give it a unique name and then supply that?  E.g.,
```
single_thread = ThreadPool(1)
shuffle_enabled = ShuffleOptions()
...
with DataLoader(Reader(..., reader_pool=single_thread, shuffle_options=shuffle_enabled, ...)?
```

examples/mnist/Dockerfile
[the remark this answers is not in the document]
forbearer: On second thought, actually, mnist generation leverages pytorch as well, so I don't think I'll gain much to introduce a pytorch subdirectory.  I'll simply add a legal header and comment about functionality to this Dockerfile.

The README.md in this directory explains use of the docker file.

examples/mnist/main.py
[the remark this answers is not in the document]
forbearer: OK, I think I see what you are getting at.  Removing these two arguments would work just fine.  :-)

PR #68, opened 2018-08-17
[Part 1] New reader flow: Epoch generator

selitvin:
Epoch generator is factored out of ventilator.py and will be used in the reader_v2


selitvin:
Found a bug. Fixing.

selitvin:
Added the unlimited epochs test. Also added a test of a multiple epochs with single item.

PR #70, opened 2018-08-18
Rework SingleFieldIndexer to support None in nullable field

meremeev:


meremeev:
how can we test for python 3.6 locally?

rgruener:
You can have a virtualenv for python3 (I have for both python2 and 3). I did need to install python3.6 though since we only have 3.4 by default on our laptops

```
virtualenv -p python3.6 env3
source env3/bin/activate
pip install -e .
py.test -v petastorm
```

PR #72, opened 2018-08-19
Try using travis-ci cache to speedup pip install phase

selitvin:
I see ~70sec goes to pip install. Try the caching feature (caches $HOME/.cache/pip

Also make pytest output verbose to see if a test is "stuck"


rgruener:
:+1:

PR #73, opened 2018-08-20
Removed no longer valid extras from .travis.yml install

selitvin:
opencv and pyarrow are no longer an extra dependencies in setup.py. Removed.


rgruener:
:+1:

PR #74, opened 2018-08-20
[Part 2] New reader: ShufflingBuffer implementations (Random and Noop)

selitvin:
Items can be added to a shuffling buffer and removed in a different order as defined by the concrete shuffling algorithm.

Currently we have a `NoopShufflingBuffer` implementation which is not really shuffling, but just a FIFO implementation. Used for tests and when we don't want to shuffle.

`RandomShufflingBuffer` implements a random shuffling buffer with low and high water marks.

Shuffling buffers are not intended to be thread safe.

### DIFF
--- a/petastorm/reader_impl/shuffling_buffer.py
+++ b/petastorm/reader_impl/shuffling_buffer.py
@@ -1,0 +1,180 @@
+#  Copyright (c) 2017-2018 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+
+import numpy as np
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class ShufflingBufferBase(object):
+    """Shuffling implements a shuffling algorithm. Items can be added to the shuffling buffer and removed in a
+    different order as defined by the concrete shuffling algorithm. A shuffling buffer is intended to be used from
+    a single thread, hence, not thread safe."""
+    @abc.abstractmethod
+    def add_many(self, items):
+        """Adds multiple items to the buffer.
+
+        :param items: items to be added to the shuffling buffer.
+        :return: None
+        """
+        pass
+
+    @abc.abstractmethod
+    def retrieve(self):
+        """Selects an item from the buffer and returns the item to the caller. The item is removed from the buffer.
+
+        :return: The selected item.
+        """
+        pass
+
+    @abc.abstractmethod
+    def can_add(self):
+        """Checks the state of the buffer and returns whether a new item can be added to the buffer at the time.
+
+        :return: A boolean indicating whether an item can be added to the buffer at the time.
+        """
+        pass
+
+    @abc.abstractmethod
+    def can_retrieve(self):
+        """Checks the state of the buffer and returns whether an item can be removed from the buffer..
+
+        :return: A boolean indicating whether an item can be returned from the buffer at the time.
+        """
+        pass
+
+    @abc.abstractproperty
+    def size(self):
+        """Returns the number of elements currently present in the buffer.
+
+        :return: number of elements currently present in the buffer
+        """
+        pass
+
+    @abc.abstractmethod
+    def finish(self):
+        """Call this method when no more `add_many` calls will be made.
+
+        This allows a user to deplete the buffer. Typically during last epoch. Otherwise, we would always have leftovers
+        in the buffer at the end of the lifecycle.
+
+        :return: number of elements currently present in the buffer
+        """
+        pass
+
+class NoopShufflingBuffer(ShufflingBufferBase):
+    """A 'no-operation' (noop) implementation of a shuffling buffer. Useful in cases where no shuffling is desired, such
+    as test scenarios or iterating over a dataset in a predeterministic order.
+    """
+    def __init__(self):
+        self.store = []
+
+    def add_many(self, items):
+        self.store.extend(items)
+
+    def retrieve(self):
+        return self.store.pop(0)
+
+    def can_retrieve(self):
+        return len(self.store) > 0
+
+    def can_add(self):
+        return True
+
+    @property
+    def size(self):
+        return len(self.store)
+
+    def finish(self):
+        pass
+
+class RandomShufflingBuffer(ShufflingBufferBase):
+    """
+    A random shuffling buffer implementation. Items can be added to the buffer and retrieved in a random order.
+    """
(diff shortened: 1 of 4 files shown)


petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+        pass
+
+class RandomShufflingBuffer(ShufflingBufferBase):
+    """
+    A random shuffling buffer implementation. Items can be added to the buffer and retrieved in a random order.
+    """
+    def __init__(self, shuffling_buffer_capacity, min_after_retrieve, extra_capacity=1000):
+        """Initializes a new ShufflingBuffer instance.
+
+        Items may be retrieved from the buffer once `min_after_dequeue` items were added to the queue
```
rgruener: min_after_dequeue -> min_after_retrieve

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Items may be added to the buffer as long as the amount of items in the buffer does not exceed
+        `shuffling_queue_capacity`.
+
+        The amount of items in the buffer may grow above `shuffling_buffer_capacity` (due to a call to `add_many`
+        with a list of items), but must remain under `extra_queue_capacity`.
+
+        :param shuffling_buffer_capacity: Items may be added to the buffer as long as the amount of items in the
+        buffer does not exceed the value of `shuffling_queue_capacity`.
+        :param min_after_retrieve: The amount of items in the buffer may grow above `shuffling_buffer_capacity`
```
rgruener: the doc for this param seems wrong

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        self._extra_capacity = extra_capacity
+        # Preallocate the shuffling buffer.
+        self._items = [None] * (shuffling_buffer_capacity + self._extra_capacity)
+        self._shuffling_queue_capacity = shuffling_buffer_capacity
+        self._min_after_dequeue = min_after_retrieve
+        self._size = 0
+        self._done_adding = False
+
+    def add_many(self, items):
```
rgruener: This doesnt look thread safe. Do we need it to be thread safe? (I think the answer is no since the main thread will "poll" and move things between queues but I wanted to be sure. Also it would be useful to document in the baseclass that a shuffling queue does not need to be thread safe)

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+                               'items can be added.')
+
+        # We leave self._extra_capacity slack to make sure we don't reallocate self._items array
+        expected_size = self.size + len(items)
+        maximal_capacity = self._shuffling_queue_capacity + self._extra_capacity
+        if expected_size > maximal_capacity:
+            raise RuntimeError('Attempt to enqueue more elements that the capacity allows. '
+                               'Current size: {}, new size {}, maximum allowed: {}'.format(self.size, expected_size,
+                                                                                           maximal_capacity))
+        self._items[self._size:len(items)] = items
```
rgruener: Shouldnt this be `self._items[self._size:self._size + len(items)] = items`

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+    def add_many(self, items):
+        if self._done_adding:
+            raise RuntimeError('Can not call add_many after done_adding() was called.')
+
+        if not self.can_add():
+            raise RuntimeError('Can not enqueue. Check the return value of "can_enqueue()" to check if more '
+                               'items can be added.')
+
+        # We leave self._extra_capacity slack to make sure we don't reallocate self._items array
+        expected_size = self.size + len(items)
```
rgruener: the mix of using `self.size` or `self._size` is a bit confusing. Perhaps stick to `self._size`

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+        """Initializes a new ShufflingBuffer instance.
+
+        Items may be retrieved from the buffer once `min_after_dequeue` items were added to the queue
+        (indicated by `can_retrieve`).
+
+        Items may be added to the buffer as long as the amount of items in the buffer does not exceed
+        `shuffling_queue_capacity`.
+
+        The amount of items in the buffer may grow above `shuffling_buffer_capacity` (due to a call to `add_many`
+        with a list of items), but must remain under `extra_queue_capacity`.
```
rgruener: maybe mention that `extra_queue_capacity` should aim to be equivalent to the maximum number of items you will try to add at once

rgruener: Would it be simpler to have chage `can_add()` to be `can_add(length)` which will return whether you can add a certain size instead of having this extra space? I can see a problem where if you have a small shuffling_buffer_capacity or try to add a lot of items at once but that shouldnt really ever be the case anyways?

petastorm/reader_impl/shuffling_buffer.py
```diff
@@ -0,0 +1,168 @@
...
+    def size(self):
+        """Returns the number of elements currently present in the buffer.
+
+        :return: number of elements currently present in the buffer
+        """
+        pass
+
+    @abc.abstractmethod
+    def finish(self):
+        """Call this method when no more `enqueu_many` calls will be made.
```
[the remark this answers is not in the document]
selitvin: Added a clarification comment:
> This allows a user to deplete the buffer. Typically during last epoch. Otherwise, we would always have leftovers in the buffer at the end of the lifecycle.

petastorm/reader_impl/shuffling_buffer.py
selitvin: done

petastorm/reader_impl/shuffling_buffer.py
selitvin: Improved the entire comment.

petastorm/reader_impl/shuffling_buffer.py
selitvin: Added a comment that this is not a threadsafe class to the base class.

petastorm/reader_impl/shuffling_buffer.py
selitvin: Checked how come the tests don't catch this. To my surprise, the second parameter does not matter! It will just use the starting point. Updated the code with your proposal since it definitely shows the intent better.

petastorm/reader_impl/shuffling_buffer.py
selitvin: Done

petastorm/reader_impl/shuffling_buffer.py
rgruener: Looking at it, I think the idiomatic way to do it would be `self._items[self._size:] = ` though I think that either way is really fine.

petastorm/reader_impl/shuffling_buffer.py
selitvin: 1. Added a clarification:
> Should be set to the upper bound of the number of items that can be added in a single call to `add_many` (can be a loose bound).

2. `can_add(length)` would be tricky as I need to know in advance if I can start pulling results from the loading queue and have a guarantee that I'll have somewhere to store them.

PR #75, opened 2018-08-20
[Part 3] New reader flow: A mock executor. Only partial implementation of `concurrent.future protocol`.

selitvin:
Guarantees stable order of future evaluations: a submitted future is evaluated immediately as part of submit implementation. Will extend this mock executor when/if needed.

### DIFF
--- a/petastorm/reader_impl/same_thread_executor.py
+++ b/petastorm/reader_impl/same_thread_executor.py
@@ -1,0 +1,29 @@
+#  Copyright (c) 2017-2018 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


petastorm/reader_impl/same_thread_executor.py
```diff
@@ -0,0 +1,58 @@
...
+    """A mock executor. Guarantees stable order of future evaluations: a submitted future is
+    evaluated immediately as part of submit implementation."""
+
+    def __init__(self):
+        # A future has to have a unique id. Use this counter to generate the id.
+        self._id = 0
+
+    def submit(self, fn, *args, **kwargs):
+        self._id += 1
+        return _SameThreadFuture(self._id, fn(*args, **kwargs))
```
rgruener: Is the SameThreadFuture class necessary? Could you not just do the following below to return futures from this exceutor?:
```
future = Future()
future.set_result(fn(*args, **kwargs))
return future
```

selitvin:
Part of the pr 100

PR #76, opened 2018-08-20
[Part 4] New reader flow: Adding a utility function that enables submitting futures to an executor using dill serializa

selitvin:


selitvin:
Apparently, it is easy to work around this issue without introducing an extra dependency on dill. Abandoning the the PR .

PR #77, opened 2018-08-20
Validate that dataset_url is string

rgruener:
Since we moved the order of arguments this validation is important to fail fast and prevent weird errors

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -15,6 +15,7 @@
 import collections
 import logging
 import os
+import six
 import warnings
 
 from pyarrow import parquet as pq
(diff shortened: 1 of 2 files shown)


petastorm/reader.py
```diff
@@ -101,6 +101,9 @@ def __init__(self, dataset_url, schema_fields=None, shuffle=None, predicate=None
         #    c. partition: used to get a subset of data for distributed training
         # 4. Create a rowgroup ventilator object
         # 5. Start workers pool
+        if dataset_url is None or not isinstance(dataset_url, str):
```
selitvin: Shouldn't it be a `basestring` (to support unicode strings)?

selitvin: Ok, `basestring` is not python3. Not sure what is the fix then, but I just saw someone calling `Reader` with a unicode path from python2.7....

selitvin: Found this is a python2,3 compatible solution:
```python
isinstance(self._dataset_url, six.string_types)
```

rgruener: Ah, yeah thats better

rgruener:
Added test

selitvin:
Sorry, should have approved last time.

PR #79, opened 2018-08-21
Make synthetic_dataset fixture shared and add an optional caching

selitvin:
Caching is controlled by `--cache-synthetic-dataset` option passed to pytest, e.g.:
```
    pytest --cache-synthetic-dataset
```
    
When specified, the `synthetic_dataset` fixture will try finding a path to the previously created dataset in pytest's cache. Otherwise, a synthetic_dataset will be recreated.


petastorm/tests/test_end_to_end.py
```diff
@@ -44,14 +42,6 @@
 SyntheticDataset = namedtuple('synthetic_dataset', ['url', 'data', 'path'])
```
rgruener: Should we not be importing this from conftest?

selitvin:
Added a https://github.com/uber/petastorm/issues/80 : track documenting this.

PR #93, opened 2018-08-24
Make `petastorm_generate_metadata.py` lunchable when not installed as an entry_point.

selitvin:
Makes it easier to run, when the package is not properly installed.


petastorm/etl/petastorm_generate_metadata.py
```diff
@@ -109,3 +109,6 @@ def _main(args):
 
 def main():
     _main(sys.argv[1:])
+
+if __name__ == '__main__':
+    _main(sys.argv[1:])
```
rgruener: I believe no newline will fail pylint

PR #99, opened 2018-08-27
Fix petastorm/__init__.py to not include predicates so pip install wo…

kashmawy:
…rks. Added development instructions to README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,3 +151,6 @@ with DataLoader(Reader('file:///localpath/mnist/test', num_epochs=10),
 # Troubleshooting
 See the [Troubleshooting](docs/troubleshooting.md) page and please submit the [ticket](https://github.com/uber/petastorm/issues/new)
 if you can't find an answer.
+
+# Development
+See the [Development](doc/development.md) page for instructions on how to develop Petastorm and run tests.
--- a/petastorm/__init__.py
+++ b/petastorm/__init__.py
@@ -12,35 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import abc
-import six
-
 __version__ = '0.3.1'
-
-
-@six.add_metaclass(abc.ABCMeta)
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -10,15 +10,15 @@ Note that although petastorm depends on opencv, it is defined as an extra packag
 to be installed by debian package (or source) as opposed to pip. It is recommended to install it when
 testing out petastorm however (and is therefore shown in the examples).  
 
-Install with tensorflow CPU dependency:
-```bash
-pip install petastorm[opencv, tf]
-```
+Please follow the following instructions to develop Petastorm:
```
selitvin: The instructions that we had here were for petastorm users, not developers (installing off pypi server). I think previous lines should still be here. Suggest we move developer related instructions to a separate `docs/development.md` (this is not a common case). 

It can be more convenient when all four lines are in the same block, so you can copy paste it at once to the terminal. We can show a safer tensorflow-cpu version and have a side comment about tf_gpu.

petastorm/predicates.py
```diff
@@ -15,12 +15,41 @@
...
+    def get_fields(self):
+        pass
+
+    @abc.abstractmethod
+    def do_include(self, values):
+        pass
+
+
+@six.add_metaclass(abc.ABCMeta)
+class RowGroupSelectorBase(object):
```
selitvin: This belongs to `selectors.py`

petastorm/__init__.py
```diff
@@ -12,35 +12,4 @@
...
 # limitations under the License.
 
-import abc
-import six
-
 __version__ = '0.3.1'
-
-
-@six.add_metaclass(abc.ABCMeta)
-class PredicateBase(object):
```
rgruener: note this is an api breaking change. We should be releasing 0.4.0 next after this change

kashmawy: yep it is.

Issue #91, opened 2018-08-23
Unify namedtuple/dict way samples are treated in Reader implementation


selitvin:
Currently we use use both namedtuple and dictionary interchangeably. This results in a lot of back and forth conversions and makes the code more cumbersome. 

Since nametuple is kinda syntactic suger for Reader users, we should make all internal usage of rows to be a dictionary (since it is easier for code parametrized by field names to handle).

PR #102, opened 2018-08-28
-Y switch and no-namedtuples-in-reader

selitvin:
Two commits in this PR:
* Adding `-Y` shortcut for `--cache-synthetic-dataset`
* Obsoleting namedtuples within the reader internals implementations. Conversion to namedtuple happens right before values are returned to the user. (Resolves https://github.com/uber/petastorm/issues/91)

### DIFF
--- a/petastorm/tests/conftest.py
+++ b/petastorm/tests/conftest.py
@@ -28,6 +28,7 @@ SyntheticDataset = namedtuple('synthetic_dataset', ['url', 'data', 'path'])
 
 ROWS_COUNT = 100
 
+_CACHE_FAKE_DATASET_OPTION_SHORT = '-Y'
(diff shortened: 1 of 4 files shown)


petastorm/tests/conftest.py
```diff
@@ -28,14 +28,15 @@
 
 ROWS_COUNT = 100
 
+_CACHE_FAKE_DATASET_OPTION_SHORT = '-Y'
```
rgruener: Just curious why `-Y` and not something like `-C` for cache?

selitvin: It was -Y for sYnthetic. Small '-c' is taken. Want to try and avoid conflicts with whatever pytest plugins we could see and -Y sounded quite isoteric.

PR #106, opened 2018-08-28
Make `ROWS_COUNT` private to conftest.py

selitvin:
No need to import it in tests. The information is present in the synthetic_dataset itself.


rgruener:
Yeah its not calling `len` on the reader but on the list of dictionaries the synthetic dataset returns

Issue #104, opened 2018-08-28
Flakey Travis CI execution of Reader tests using ThreadPool


forbearer:
A couple examples of such run are:
* Python 2.x: https://travis-ci.com/uber/petastorm/jobs/142290679
    * I've seen the issue happen more regularly with 2.x
* Python 3.x: https://travis-ci.com/uber/petastorm/jobs/142290686

```
petastorm/tests/test_end_to_end.py::test_simple_read[reader_factory1] /home/travis/.travis/job_stages: line 78:  2987 Segmentation fault      (core dumped) pytest -v --ignore=petastorm/tests/test_pytorch_utils.py --ignore=examples/mnist/tests/test_generate_mnist_dataset.py --cov=./
```

forbearer:
@selitvin noted that, despite the ignore, the effect of `import pytorch` in `examples/mnist/tests/conftest.py` must still be bleeding into the main pytest run.

It turns out that that makes sense, because pytest will read the conftest as part of plugin scan.

So, this `-p no:<plugin>` *should* help us!

forbearer:
An alternative approach (which I thought I had tried without success once before, but re-visiting, and manipulating order of the tests as well):  Do `import torch` early as possible in test_generate_mnist_dataset.py, and make sure that's ordered first when running pytest.  This way, I've not introduced a plugin that could bleed `pytorch` module into other tests.

Let's see how this run goes: https://travis-ci.com/uber/petastorm/jobs/142703106

selitvin:
Resolved by adding `import pyarrow` before loading `pytorch` on all code paths.

PR #107, opened 2018-08-28
Install gdb and enable core dump

forbearer:
In service of uber/petastorm#104...need to install `gdb` and enable core!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,17 @@ install:
   # (or optionally install pyarrow from source instead of through binaries)
   - pip install --upgrade numpy
 
+before_script:
+  # enable core dumps
+  - ulimit -c unlimited -S
+
 script:
   - pylint --rcfile=.pylintrc petastorm examples -f parseable -r n
   # Completely ignore two pytorch tests to prevent static-TLS-caused torch-import seg-fault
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,6 +19,10 @@ python:
   - '2.7'
   - '3.6'
 
+before_install:
+  # Support processing of core dump
+  - sudo apt-get install -y gdb
```
rgruener: As discussed in person, we can probably install gdb in the after failure so that it only gets installed when we actually need it

forbearer:
Experimented with the following, but unsurprisingly, mixing-and-match executable don't work. :P

```
after_failure:
  # Only upon failure, install gdb to process core dump
  - sudo apt-get install -y gdb python2.7-dbg
  - COREFILE=$(find . -maxdepth 2 -name "core*" | head -n 1)	  - COREFILE=$(find . -maxdepth 2 -name "core*" | head -n 1)
  - if [[ -f "$COREFILE" ]]; then gdb -c "$COREFILE" pytest -ex "thread apply all bt" -ex "set pagination 0" -batch; fi	  - if [[ -f "$COREFILE" ]]; then gdb -c "$COREFILE" python2.7-dbg -ex "thread apply all bt" -ex "set pagination 0" -batch; fi
```

forbearer:
I've confirmed that this setup processes the core file from pytest run that dumps core, but there's no useful bt output without debug information.

PR #108, opened 2018-08-28
Removing early loading of pytorch.

selitvin:
Let's see if this helps CI builds pass.

It was leaking into non pytorch tests and triggering a crash of the tests.
Still need to figure out why the crash happens. a


forbearer:
Your test confirms what I saw as well, that the pytorch tests now fails (`pytest -v examples/mnist/tests/test_generate_mnist_dataset.py petastorm/tests/test_pytorch_utils.py`).

I am surprised that `import pytorch` in this particular conftest should leak elsewhere, especially since the test in that path is ignored.

Ah, what if we explicitly ignore this conftest file as well?

selitvin:
@forbearer , is the purpose of the import in the conftest.py to prevent TLS errors? Is it still the case and we need it for that purpose?

forbearer:
Yep, that is the reason.

Without this import, the dlopen static TLS error causes the pytest collect to seg fault.

forbearer:
Let's land this for now, plz

PR #110, opened 2018-08-29
Use absolute path to ignore plugin containing pytorch import

forbearer:
Still in service of #104


forbearer:
According to this [latest CI run](https://travis-ci.com/uber/petastorm/jobs/142679227#L577), clearly `examples/mnist/tests/conftest.py` is ignored from the plugin list, so why is the test run still failing at the same point?!

PR #114, opened 2018-08-30
Sphinx autodoc and integration into readthedoc.org

forbearer:
This only contains a very partial list, but I wanted to share how this looks and get some initial feedback on whether we like this or not.  Unfortunately, I don't think there's a way to view the index.html files as HTML directly via github links...so may need to pull the branch, then navigate to `petastorm/docs/_build/html/index.html` to see the doc.

In parallel, I'm experimenting with [ReadTheDocs hook into github](https://docs.readthedocs.io/en/latest/getting_started.html#sign-up-and-connect-an-external-account) to see what that looks like.

Here's an experimental project link: https://petastorm.readthedocs.io/en/latest/

Given that petastorm is now open source, should there be any concerns about the documentation setup and linkage to Readthedocs?

(EDIT: It turns out none of the `_builds` artifacts needs to be checked in, as readthedocs builds everything. Also, @rgruener found that we have a corp shared account for readthedocs!)


petastorm/unischema.py
```diff
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""'unischema' is a data structure definition which can be rendered as native schema/data-types objects
-in several different python libraries. Currently supported arepyspark, tensorflow, numpy"""
+"""A ``unischema`` is a data structure definition which can be rendered as native schema/data-types objects
```
forbearer: The changes in this and codec.py show in https://petastorm.readthedocs.io/en/latest/data_schema.html .

petastorm/codecs.py
```diff
@@ -27,6 +29,7 @@
```
forbearer: Unfortunately, the sphinx build is unable to deal with `cv2` import without blowing memory limitations (free service... apparently the boxes only get 1 GB RAM).  So, `codec` won't get processed.

One alternative is if `import cv2` happens only within `CompressedImageCodec.__init__`?! :D

rgruener: Another option is to just not import it when doing a read the docs? https://docs.readthedocs.io/en/latest/faq.html#how-do-i-change-behavior-for-read-the-docs

docs/extensions.rst
```diff
@@ -0,0 +1,23 @@
+Extensions
```
selitvin: An idea: maybe we hide these 'frameworkie' kind of files that have little value for browsing on github in a separate subdirectory (docs/autodoc or something?). That way, it would be easier to see the development.md and other browsable files easier?

.gitignore
```diff
@@ -5,3 +5,5 @@
 petastorm.egg-info
 build
 dist
+
+_build
```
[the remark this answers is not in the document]
forbearer: for sure, will do!

docs/extensions.rst
forbearer: sphinx autodoc assumes a `docs` directory, but that is configurable.

That said, if we fold the .md files into readthedocs, then all files under `docs/` would become the same "type", right?

forbearer: @selitvin Would it be OK if I address this in a separate PR where I make all .mds rst files?

petastorm/codecs.py
forbearer: So, either way, the `codecs.py` would simply to skipped. It doesn't cause the entire autodoc build process to fail.

forbearer:
@selitvin I've added steps for doc generation to `development.md`. If it looks good to you, I'll merge this PR and then proceed to do the next things.

forbearer:
Good questions!

The current setup isn’t final yet, but once I merge, I can update the RTD
petastorm account to point to master.

There is the notion of `latest` and `stable`, as well as any branch and
release based versions.

It appears that we can configure `latest` to point to one of these; that’s
how I have it for now, pointed to my fork on the `doc_autogen` branch.  And
thus petastorm.rtfd.io updates upon push to the branch.

We can pick master, or once we cut a release containing this PR, that new
release as latest. That would ensure we only change the live autodoc when
we choose to.

Does that clarify?  I’ll make this clear in the doc as well.

PR #115, opened 2018-08-30
Add deploy to public pypi

rgruener:


setup.cfg
```diff
@@ -0,0 +1,2 @@
+[bdist_wheel]
+universal = 1
```
rgruener: Since all our code is python2/3 compatible we can actually make a single universal wheel. I have tested this in rAV and it seems to work fine

README.md
```diff
@@ -149,8 +149,8 @@ with DataLoader(Reader('file:///localpath/mnist/test', num_epochs=10),
 
 
 # Troubleshooting
-See the [Troubleshooting](docs/troubleshoot.md) page and please submit a [ticket](https://github.com/uber/petastorm/issues/new)
+See the [Troubleshooting](https://github.com/uber/petastorm/blob/master/docs/troubleshoot.md) page and please submit a [ticket](https://github.com/uber/petastorm/issues/new)
```
rgruener: This was changed so that these links work on pypi.org

rgruener:
To truly test, we will need to wait until we actually perform a release :smile:

PR #119, opened 2018-09-04
add compression option to NdarrayCodec() to support Q10.8 encoded fie…

meremeev:
…lds with compression

### DIFF
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -128,6 +128,36 @@ class NdarrayCodec(DataframeColumnCodec):
         return BinaryType()
 
 
+class CompressedNdarrayCodec(DataframeColumnCodec):
+    """Encodes numpy ndarray with compression into a spark dataframe field"""
+
+    def encode(self, unischema_field, array):
+        expected_dtype = unischema_field.numpy_dtype
+        if isinstance(array, np.ndarray):
+            if expected_dtype != array.dtype.type:
+                raise ValueError('Unexpected type of {} feature. '
+                                 'Expected {}. Got {}'.format(unischema_field.name, expected_dtype, array.dtype))
+
+            expected_shape = unischema_field.shape
+            if not _is_compliant_shape(array.shape, expected_shape):
+                raise ValueError('Unexpected dimensions of {} feature. '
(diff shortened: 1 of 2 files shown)


petastorm/codecs.py
```diff
@@ -101,6 +101,9 @@ def spark_dtype(self):
 class NdarrayCodec(DataframeColumnCodec):
     """Encodes numpy ndarray into a spark dataframe field"""
 
+    def __init__(self, compression=False):
+        self.compression = compression
```
selitvin: Doc string please.

petastorm/tests/test_codecs.py
```diff
@@ -34,6 +34,16 @@ def test_numpy_codec(self):
                                nullable=False)
         np.testing.assert_equal(codec.decode(field, codec.encode(field, expected)), expected)
 
+class NumpyArrayCodecsCompressionTest(unittest.TestCase):
+
+    def test_numpy_codec(self):
+        SHAPE = (10, 20, 30)
+        expected = np.random.rand(*SHAPE).astype(dtype=np.int32)
+        codec = NdarrayCodec(compression=True)
+        field = UnischemaField(name='test_name', numpy_dtype=np.int32, shape=SHAPE, codec=NdarrayCodec(),
```
selitvin: Should we use the same codec instead of creating a new instance? (`codec=NdarrayCodec()`)

selitvin:
@rgruener , do you know if there is a way to specify a native parqurt compression algorithm on a per column basis when writing out from spark?

rgruener:
Unfortunately that is not possible, compression is specified on the entire output.

petastorm/tests/test_codecs.py
meremeev: It is the same codec. test check compression option. Codec in UnischemaField is not used, exactly as in original test above.

petastorm/codecs.py
meremeev: done

selitvin: My question is whether you can read existing datasets with this code (since the signature of the codec has changes). Can you please confirm against msds 17 for example?

PR #120, opened 2018-09-05
Added tensorflow_example. Updated the README and moved main.py to pytorch_example.py

kashmawy:


examples/mnist/tf_example.py
```diff
@@ -0,0 +1,125 @@
...
+
+    sess = tf.InteractiveSession()
+    tf.global_variables_initializer().run()
+
+    # Train
+    print("Training model for {0} epochs with batch size {1} and log interval {2}".format(
+        epochs, batch_size, log_interval
+    ))
+    for i in range(epochs):
+        batch_x, batch_y = get_batch(train_reader, batch_size)
```
selitvin: Suggest we show our TF native interface:`tf_tensors` together with a standard `tf.train.batch`for batching. Here you use pure python API which is less interesting in our context (we don't want to show how to train MNIST, but to show how we hookup reader efficiently to train MNIST).

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -18,7 +18,7 @@
 import sys
 import torch  # pylint: disable=unused-import
 
-import examples.mnist.main as main
+import examples.mnist.pytorch_example as pytorch_example
```
forbearer: Thank you for this rename, @kashmawy.

One more thing, did you make sure to run the test to ensure passing?\
1. `cd examples/mnist/tests`
2. `PYTHONPATH=<path_to_repo> pytest -v`

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,137 @@
...
+                ])
+                coord = tf.train.Coordinator()
+                threads = tf.train.start_queue_runners(sess=sess, coord=coord)
+                try:
+                    for i in range(epochs):
+                        if coord.should_stop():
+                            break
+
+                        feed_image_batch, feed_label_batch = sess.run([train_batch_image, train_batch_label])
+                        sess.run(train_step, feed_dict={x: feed_image_batch, y_: feed_label_batch})
```
selitvin: You should connect everything in a graph - not run part of the graph to get the data and then use feed dict.
Can you do something like this:
```            
cross_entropy = tf.losses.sparse_softmax_cross_entropy(labels=feed_label_batch, logits=feed_image_batch)
```
and then evaluate the training step without feeddict?
```
sess.run(train_step)
```

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,137 @@
...
+                             '(default: %s)' % default_dataset_url)
+    parser.add_argument('--epochs', type=int, default=100, metavar='N',
+                        help='number of epochs to train (default: 100)')
+    parser.add_argument('--batch-size', type=int, default=100, metavar='N',
+                        help='input batch size for training (default: 100)')
+    parser.add_argument('--log-interval', type=int, default=10, metavar='N',
+                        help='how many batches to wait before logging training status (default: 10)')
+    args = parser.parse_args()
+
+    train_and_test(
```
selitvin: Can you also please cover this with a test (same as what was done for pytorch). This will make sure the code is not getting stale.

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,137 @@
...
+
+            with tf.Session() as sess:
+                sess.run([
+                    tf.local_variables_initializer(),
+                    tf.global_variables_initializer(),
+                ])
+                coord = tf.train.Coordinator()
+                threads = tf.train.start_queue_runners(sess=sess, coord=coord)
+                try:
+                    for i in range(epochs):
```
selitvin: Epochs is a number of full iterations over your dataset. This is not the case with your code. Note that `Reader` will never produce more then 1 epochs with this code.

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,137 @@
...
+                            feed_image_batch, feed_label_batch = sess.run([test_batch_image, test_batch_label])
+                            print('After training for {0} epochs, the accuracy of the model is: {1:.2f}'.format(
+                                i,
+                                sess.run(accuracy, feed_dict={x: feed_image_batch, y_: feed_label_batch})))
+                finally:
+                    coord.request_stop()
+                    coord.join(threads)
+
+def transform(readout_example):
+    image = tf.cast(tf.reshape(readout_example.image, [784]), tf.float32)
```
selitvin: Consider removing this function all together and just do it inline. It does not do much and the name does help understand its functionality.

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,130 @@
...
+                try:
+                    for i in range(training_iterations):
+                        if coord.should_stop():
+                            break
+
+                        sess.run(train_step)
+
+                        if (i % log_interval) == 0 or i == (training_iterations - 1):
+                            correct_prediction = tf.equal(tf.argmax(y, 1), batch_label)
+                            accuracy = tf.reduce_mean(tf.cast(correct_prediction, tf.float32))
```
selitvin: This extends TF graph everytime you go through this code. Eventually you run out of memory. Can you see if you can do the entire graph construction before the train eval loop?

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,130 @@
...
+    parser.add_argument('--dataset-url', type=str,
+                        default=default_dataset_url, metavar='S',
+                        help='hdfs:// or file:/// URL to the MNIST petastorm dataset'
+                             '(default: %s)' % default_dataset_url)
+    parser.add_argument('--training-iterations', type=int, default=100, metavar='N',
+                        help='number of training iterations to train (default: 100)')
+    parser.add_argument('--batch-size', type=int, default=100, metavar='N',
+                        help='input batch size for training (default: 100)')
+    parser.add_argument('--log-interval', type=int, default=10, metavar='N',
+                        help='how many batches to wait before logging training status (default: 10)')
```
selitvin: Logging : maybe evaluation describes the action better?

examples/mnist/tests/test_generate_mnist_dataset.py
```diff
@@ -140,17 +142,31 @@ def test_full_pytorch_example(large_mock_mnist_data, tmpdir):
...
+    mnist_data_to_petastorm_dataset(tmpdir, dataset_url, mnist_data=large_mock_mnist_data,
+                                    spark_master='local[1]', parquet_files_count=1)
+
+    # Tensorflow train and test
+    tf_example.train_and_test(
+        dataset_url=dataset_url,
+        training_iterations=10,
+        batch_size=10,
+        evaluation_interval=10,
+    )
```
selitvin: Missing EOL. Please run linter (see `.travis.yml` for the exact command)

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,130 @@
...
+                threads = tf.train.start_queue_runners(sess=sess, coord=coord)
+                try:
+                    for i in range(training_iterations):
+                        if coord.should_stop():
+                            break
+
+                        sess.run(train_step)
+
+                        if (i % evaluation_interval) == 0 or i == (training_iterations - 1):
+                            feed_batch_image, feed_batch_label = sess.run([test_batch_image, test_batch_label])
```
selitvin: Can this be wired directly into the evaluation portion of the graph? Same as what you have done before with the training part?

kashmawy: As discussed, this is not possible, because we need to feed test data during training to the same tensors.

examples/mnist/tf_example.py
```diff
@@ -0,0 +1,130 @@
...
+import tensorflow as tf
+
+from examples.mnist import DEFAULT_MNIST_DATA_PATH
+from petastorm.reader import Reader
+from petastorm.tf_utils import tf_tensors
+
+
+def train_and_test(dataset_url, training_iterations, batch_size, evaluation_interval):
+    """
+    Train a model for epochs with a batch size batch_size, printing accuracy every log_interval.
```
selitvin: These are not epochs.

PR #121, opened 2018-09-06
add shuffle method to tf_utils

jakecharland:
This change is based off a diff @selitvin made for me to enable mixin petastorm readers. Some things have changed in tf_utils since the original change was made and I updated as best I could with my limited knowledge of the tf_utils code.

### DIFF
--- a/petastorm/tf_utils.py
+++ b/petastorm/tf_utils.py
@@ -310,3 +310,14 @@ def  tf_tensors(reader, shuffling_queue_capacity=0, min_after_dequeue=0):
         result = _tf_tensors_nonngram(reader, shuffling_queue_capacity, min_after_dequeue)
 
     return result
+
+def shuffle(reader, shuffling_queue_capacity, min_after_dequeue, vals):
+    """ Creates a TensorFlow shuffling queue using the specified capacity and reader.
+    """
+    shuffled_tensors = _shuffling_queue(shuffling_queue_capacity, min_after_dequeue, [v.dtype for v in vals], vals)
+    if reader.ngram:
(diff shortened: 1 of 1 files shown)


jakecharland:
Would it make sense to add the mixin reader support to the petastorm reader? It could function by accepting a list of datasets and a list of sample probability.

petastorm/tf_utils.py
```diff
@@ -310,3 +310,14 @@ def  tf_tensors(reader, shuffling_queue_capacity=0, min_after_dequeue=0):
         result = _tf_tensors_nonngram(reader, shuffling_queue_capacity, min_after_dequeue)
 
     return result
+
+def shuffle(reader, shuffling_queue_capacity, min_after_dequeue, vals):
```
rgruener: Can you add params to docstring and maybe give example usage?

rgruener:
> Would it make sense to add the mixin reader support to the petastorm reader?
Yes, I think that would make sense and is a feature the base reader should be able to support

selitvin:
I wonder if it would make more sense to have a puthon object that would mix two readers and have a reader interface? That way, the mixing would also work for patch.

jakecharland:
Ya I guess my impl would not work here as it is tensorflow specific.

selitvin:
@jakecharland , 
Do you want to create the python version of this mixing algorithm? It can be used for stratified sampling, which is a requested feature...  

An idea could be an object that complies with `Reader` interface (an iterable context manager), takes a list of Readers and probabilities.

selitvin:
Superseded by PR 137

PR #123, opened 2018-09-06
Refactoring of documentation to rST and reorg autodocs

forbearer:
Here's what the updated rtfd result looks: https://petastorm.readthedocs.io/en/docs_refactor/

Key changes of note:
* Sphinx auto doc support files got moved into `docs/autodoc/`
* README and 3 markdown files under `docs/` are converted to rST files. Let's make sure we'll be happy with this going forward...or, we can opt to rollback on this attempt.

Oh, I just realized there is at least one README.md file missed in the repo hierarchy!


forbearer:
I redid my commits so that the 4 file renames are done first. This way, a reviewer can look at just the MD->RST commit to look at changes for the 4 .md files (e.g., https://github.com/uber/petastorm/pull/123/commits/2513a99ab687808c383678b03b367b564ec6f48a ).

docs/autodoc/reader.rst
```diff
@@ -1,51 +1,51 @@
-The Reader API
-==============
+``reader``
```
selitvin: This now looks like code ("`"). As a reader, I would expect to find corresponding code elements (packages or classes). Should't it be the same style as before?

forbearer: Oh yeah, for side by side comparison:
* Master: https://petastorm.readthedocs.io/en/latest/
* New look: https://petastorm.readthedocs.io/en/docs_refactor/

selitvin: In the example you refer to these are the package names, hence it makes sense to use this style. In our case, it's just 'headings'. We should improve our package structure so they correspond, I agree. But for now, IMO it's a little confusing.

forbearer: Ah, that's a valid point.  I may have gone too far in the name of consistency.

Shall I restore the original non-code-block headings?

forbearer: After Slack convo, we concluded that the original heading style is better for now.

So I've restored the heading style and pushed that.

PR #128, opened 2018-09-07
An example of reading a dataset from pyspark

selitvin:
Includes three commits:
- Simpler way of getting a unischema from a dataset by url
- Remove `._common_metadata.crc` file, so a locally created dataset is readable by spark
- A hello world example : using pyspark to read from a dataset.


petastorm/utils.py
```diff
@@ -89,3 +94,15 @@ def add_to_dataset_metadata(dataset, key, value):
...
     with dataset.fs.open(common_metadata_file_path, 'wb') as metadata_file:
         pyarrow.parquet.write_metadata(schema, metadata_file)
+
+    # We have just modified _common_metadata file, but the filesystem implementation used by pyarrow does not
+    # update the .crc value. We better delete the .crc to make sure there is no mismatch between _common_metadata
+    # content and the checksum.
+    if isinstance(dataset.fs, LocalFileSystem) and dataset.fs.exists(common_metadata_file_crc_path):
+        try:
+            dataset.fs.rm(common_metadata_file_crc_path)
+            logger.warn('Appears that pyarrow.filesystem.LocalFileSystem now supports `rm` method. '
```
rgruener: Im not sure we should be using this warning to alert us this can be removed since it is a burden on users and not on us as maintainers.

selitvin: A good point.

PR #129, opened 2018-09-07
Test case added to fix #43, but found that the error doesn't occur

### LINKED ISSUE #43: Querying a single column with a predicate on the same column will not return any results
The following pseudo code would fail:
```python
reader = Reader([SomeSchema.a], predicate=in_lambda([SomeSchema.a], lambda a: True),...)
reader.fetch()
```
Will never find results.

I think the reason is due to this piece of code in the Reader.py:
```python
# Read remaining columns
other_rows = piece.read(
    open_file_func=lambda filepath: file,
    columns=other_column_names,
    partitions=self._dataset.partitions).to_pandas().to_dict('records')

# Remove rows that were filtered out by the predicate
filtered_other_rows = [row for i, row in enumerate(other_rows) if match_predicate_mask[i]]

# Decode remaining columns
decoded_other_rows = [_decode_row(row, self._schema) for row in filtered_other_rows]

# Merge predicate needed columns with the remaining
all_cols = [_merge_two_dicts(a, b) for a, b in zip(decoded_other_rows, filtered_decoded_predicate_rows)]
```
`other_rows` will be `[]` since no columns were loaded. This will result in `zip(decoded_other_rows, filtered_decoded_predicate_rows)` returning an empty list `[]` even `filtered_decoded_predicate_rows` could be an array of `[True, True,....]`

forbearer:
It's possible that I didn't fully grok #43, but the newly added test case seems to pass.

Perhaps someone already fixed #43?!

But, this pull request is a refactoring of `test_predicates.py` to pytest form...please use diff view ignoring spaces. :-)

### DIFF
--- a/petastorm/tests/test_predicates.py
+++ b/petastorm/tests/test_predicates.py
@@ -12,102 +12,117 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import unittest
+import pytest
 
 from petastorm.predicates import in_set, in_intersection, \
     in_negate, in_reduce, in_pseudorandom_split, in_lambda
+from petastorm.reader import Reader
+from petastorm.tests.test_common import TestSchema
+from petastorm.workers_pool.dummy_pool import DummyPool
 
 
-class PredicatesTest(unittest.TestCase):
+@pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


selitvin:
Nice! This helps.

PR #130, opened 2018-09-07
Another pyspark example showing an SQL query.

selitvin:


selitvin:
Good point. Will do: https://github.com/uber/petastorm/issues/132

PR #133, opened 2018-09-10
A standard palette to be used throught the documentation

selitvin:


selitvin:
I used this palette in the eng blog post for the figures. You are welcome to give it a shot. The palette is not set in stone, we can change it if you think we need to swap a color or two.

PR #135, opened 2018-09-10
Speed up processing of small row datasets with ProcessPool

selitvin:
We use pyarrow serialization instead of standard pickling. Unfortunatelly, pyarrow serialization
does not support decimals, so we modify the codec to decode decimals always as strings.

Updated all tests to reflect the decimal->string behavior.


selitvin:
Added documentation to codecs.py, although it is not very visible. Will need to work on a better user guide.

PR #137, opened 2018-09-11
Adding WeightedSamplingReader class

selitvin:
`WeightedSamplingReader` takes a list of Reader objects and probabilities. When iterated over, it returns a mixture of rows returned from the readers. The mixture proportions defined by the probabilities passed to `WeightedSamplingReader` in a constructor.
`WeightedSamplingReader`  implements the same interface as `Reader`.

### DIFF
--- a/petastorm/sampling_mixer.py
+++ b/petastorm/sampling_mixer.py
@@ -1,0 +1,80 @@
+#  Copyright (c) 2017-2018 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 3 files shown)


petastorm/sampling_mixer.py
```diff
@@ -0,0 +1,80 @@
...
+        return next(self._readers[reader_index])
+
+    def next(self):
+        return self.__next__()
+
+    # Functions needed to treat reader as a context manager
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
rgruener: I am slightly concerned if we ever have a reader implementation that does more than just `stop` and `join` in the exit but I dont think theres a good way to wrap context manager functions so I think this is fine actually

PR #140, opened 2018-09-12
Revert to using the old metadata method by default

rgruener:
Since we have needed to do this a few times, lets bring the legacy metadata back for now until we can figure it out more explicitly

### DIFF
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -15,7 +15,6 @@
 import json
 import logging
 import os
-import pickle
 from contextlib import contextmanager
 from operator import attrgetter
 
@@ -32,7 +31,15 @@ ROW_GROUPS_PER_FILE_KEY = b'dataset-toolkit.num_row_groups_per_file.v1'
 UNISCHEMA_KEY = b'dataset-toolkit.unischema.v1'
 
 
-class MetadataGenerationError(Exception):
+class PetastormMetadataError(Exception):
+    """
+    Error to specify when the petastorm metadata does not exist, does not contain the necessary information,
+    or is corrupt/invalid.
+    """
+    pass
+
+
+class PetastormMetadataGenerationError(Exception):
     """
     Error to specify when petastorm could not generate metadata properly.
     This error is usually accompanied with a message to try to regenerate dataset metadata.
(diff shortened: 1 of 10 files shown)


petastorm/etl/dataset_metadata.py
```diff
@@ -77,7 +76,10 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None):
         validate_schema=False)
 
     _generate_unischema_metadata(dataset, schema)
-    if not dataset.metadata_path:
+    try:
```
rgruener: Another option here is to check if the _metadata file was generated and automatically using the legacy method if not. I think its best to avoid that for now though unless it becomes a real problem.

petastorm/etl/dataset_metadata.py
```diff
@@ -77,7 +76,10 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None):
...
+        # Try to load the row groups, if it fails that means the metadata was not generated properly
+        load_row_groups(dataset)
+    except ValueError:
```
selitvin: A little strange that you catch `ValueError` here, as it indicates invalid arguments to the called function. I think we are trying to communicate something more concrete. Maybe a custom error class for this case is more appropriate?

petastorm/etl/dataset_metadata.py
```diff
@@ -153,7 +155,7 @@ def load_row_groups(dataset):
     :return: splitted pieces, one piece per row group
     """
     # Split the dataset pieces by row group
-    metadata = dataset.metadata
+    metadata = dataset.metadata or dataset.common_metadata
```
selitvin: Maybe comment why do we do the `or` here.

petastorm/etl/dataset_metadata.py
```diff
@@ -146,35 +167,63 @@ def _generate_unischema_metadata(dataset, schema):
...
+        raise ValueError('Expected dataset.paths to be a single path, not a list of paths')
+
+    # Get the common prefix of all the base path in order to retrieve a relative path
+    paths = [piece.path for piece in dataset.pieces]
+
+    # Needed pieces from the dataset must be extracted for spark because the dataset object is not serializable
+    fs = dataset.fs
+    base_path = dataset.paths
+    row_groups = spark_context.parallelize(paths, len(paths)) \
+        .map(lambda path: (os.path.relpath(path, base_path), pq.read_metadata(fs.open(path)).num_row_groups)) \
```
selitvin: Consider explicitly closign the file.

PR #160, opened 2018-09-15
Disable a flavor of reader that is suspected to hang unit tests on travis

selitvin:


selitvin:
Probably 162 has a better solution.

PR #170, opened 2018-09-20
Initial benchmarking implementation

selitvin:
Split into multiple commits for easier review.


docs/benchmarks_tutorial.rst
```diff
@@ -0,0 +1,42 @@
...
+Now, let's run ``petastorm-throughput``:
+
+.. code-block:: bash
+
+ $ petastorm-throughput.py file:///tmp/hello_world_dataset
+ Average sample read rate: 709.84 samples/sec; RAM 217.12 MB (rss); CPU 136.30%
+
+Typically, you would want to find an optimal throughput by varying the following parameters:
+
+ - ``-w``: number of workers (controls number of loaders in the experimental :class:`~petastorm.reader.reader_v2.ReaderV2`).
```
rgruener: I think its a bit unclear the difference between number of loaders and thread based parallelism here

petastorm/benchmark/cli.py
```diff
@@ -0,0 +1,123 @@
...
+    logger.info('Done')
+    print('Average sample read rate: {:1.2f} samples/sec; RAM {:1.2f} MB (rss); '
+          'CPU {:1.2f}%'.format(results.samples_per_second, results.memory_info.rss / 2**20, results.cpu))
+
+
+def main():
+    _main(sys.argv[1:])
+
+
+if __name__ == '__main__':
```
rgruener: The order of commits seems a bit off since this file was referenced in earlier commits. This isnt a huge deal but could cause trouble if we were to revert based on commit (which we likely wont anyways)

docs/benchmarks_tutorial.rst
selitvin: Clarified a little. Once we get rid of V1, we'll cleanup the documentation.

PR #172, opened 2018-09-21
Add publications link to eng doc and add logo to RTD navigation pane

forbearer:
Also:
* moved the inclusion marker to ensure the pyspark and SQL examples show up in RTD.
* renamed autodoc content from from `reader.rst` to `api.rst`. :-)


selitvin:
image
It looks a little big to me. Also, we should remove the right logo from the main page. Two logos seems a bit too much :)

forbearer:
Haha.  I'm happy to remove the right logo.

As for the left logo, I don't see a way to tune the size in the alabaster theme, which means I'll need to make a new scaled-down logo.

We have 2 logo files right now, the smaller at 120.  I shall make a 75x75 and see?

forbearer:
So, here's what it looks like with a 75x70 px logo in the navigation pane, and no logo right-side.  I know it looks off-centered, but from the panel box, it's actually exactly centered.  It'll be more difficult to try to get that exactly right.  Let me know your thoughts:

image

forbearer:
Is it the centering? Or anything else you’d like me to try, Yevgeni?

selitvin:
It does look misaligned right now. Not sure how to make it better. Suggest we land it as is for now.

forbearer:
Ok. Will do.

If I have time. I’ll see if there’s a reasonable CSS customization to make
the alignment visually more pleasing.

Issue #178, opened 2018-09-23
Reduce code complexity of worker_loop function


selitvin:
Currently, flake8 fails on worker_loop being too complex function. Should evaluate if the function can be refactored into smaller pieces.
The following check is currently failing:
```
flake8 . --count --max-complexity=10 --statistics
```
Enable the check as a blocking once the code is refactor and the test passes.

cclauss:
https://github.com/uber/petastorm/blob/master/petastorm/reader_impl/worker_loop.py#L52 is too complex.

PR #179, opened 2018-09-23
Making a wider range of flake8 checks be blocking

selitvin:
On the way moving pylint into before_script section to fail faster on linting.

@cclauss , added you as a collaborator, if you are interested in reviewing this PR.


selitvin:
The second run is non-blocking on purpose, until #178 is resolved. Will make both runs into one, once it is resolved, as per instructions in #178.

PR #182, opened 2018-09-25
A command line copy dataset utility

selitvin:
A command line utility that copies a dataset while selecting a subset of columns/rows.

### DIFF
--- a/petastorm/tools/copy_dataset.py
+++ b/petastorm/tools/copy_dataset.py
@@ -1,0 +1,142 @@
+#  Copyright (c) 2017-2018 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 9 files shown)


petastorm/tools/copy_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+    if field_regex and not fields:
+        field_names = list(schema.fields.keys())
+        raise ValueError('Regular expressions (%s) do not match any fields (%s)', str(field_regex), str(field_names))
+
+    if fields:
+        subschema = schema.create_schema_view(fields)
+    else:
+        subschema = schema
+
+    with materialize_dataset(spark, target_url, subschema, 196):
```
rgruener: row group size should probably be configurable by the command line tool

petastorm/tools/copy_dataset.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def _main(sys_argv):
+    logging.basicConfig()
+
+    args = args_parser().parse_args(sys_argv)
+
+    spark = configure_spark(SparkSession.builder.appName('petastorm-copy'), args).getOrCreate()
+
+    spark.conf.set('spark.sql.files.maxPartitionBytes', '1010612736')
```
rgruener: add comment about this option being set due to large row sizes

PR #184, opened 2018-09-26
Add option to use custom pyarrow filesystem when materializing datasets

rgruener:


rgruener:
Lionel from maps who is using `libhdfs` also cannot use libhdfs3 when running spark and generating the dataset.

Added a test

PR #185, opened 2018-09-26
Batch decoding tasks for decoding. Optionally use `pyarrow.serialize`

selitvin:
This is an optimization that helps in case of a small row sizes and decoders running in a process pool.
When bundled, a decoder returns multiple rows in a single queued message and we potentially save tens of thousands of small message passed cross process boundaries.

### DIFF
--- a/petastorm/reader_impl/reader_v2.py
+++ b/petastorm/reader_impl/reader_v2.py
@@ -16,14 +16,16 @@ import logging
 import os
 import threading
 from collections import Counter
-
-import six
 from concurrent.futures import ThreadPoolExecutor
+
+import pyarrow
+import six
 from pyarrow import parquet as pq
 from six.moves.queue import Queue
 from six.moves.urllib.parse import urlparse
 
 from petastorm.cache import NullCache
+from petastorm.codecs import decimal_to_str
 from petastorm.etl import dataset_metadata, rowgroup_indexing
 from petastorm.fs_utils import FilesystemResolver
 from petastorm.ngram import NGram
@@ -38,7 +40,18 @@ from petastorm.shuffle_options import ShuffleOptions
 
 logger = logging.getLogger(__name__)
 
-_OUTPUT_QUEUE_SIZE = 30
+
+class PyArrowSerializeRowGroupSerializer(object):
+
+    @staticmethod
+    def serialize(rows):
+        # pyarrow.serialize does not support decimals.
+        decimal_to_str(rows)
+        return pyarrow.serialize(rows).to_buffer()
+
+    @staticmethod
+    def deserialize(serialized_rows):
+        return pyarrow.read_serialized(serialized_rows).deserialize()
 
 
 class ReaderV2(object):
@@ -47,7 +60,7 @@ class ReaderV2(object):
     def __init__(self, dataset_url, schema_fields=None, shuffle=None, predicate=None, rowgroup_selector=None,
                  num_epochs=1, sequence=None, training_partition=None, num_training_partitions=None,
                  read_timeout_s=None, cache=None, loader_pool=None, decoder_pool=None, shuffling_queue=None,
-                 shuffle_options=None, pyarrow_filesystem=None):
+                 shuffle_options=None, pyarrow_filesystem=None, pyarrow_serialize=False):
         """Initializes a reader object.
 
         :param dataset_url: an filepath or a url to a parquet directory,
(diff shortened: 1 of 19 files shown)


petastorm/reader_impl/worker_loop.py
```diff
@@ -86,8 +129,7 @@ def worker_loop(epochs_generator, loading_pool, loader, decoding_pool, decoder,
             # infinite epochs.
             if rowgroup_spec is not None and len(in_loading_futures) < _MAX_IN_LOADING_QUEUE_SIZE:
                 in_loading_futures.append(loading_pool.submit(loader.load, rowgroup_spec))
-                stats['rowgroups_scheduled_for_loading'] += 1
-                stats['in_loading_queue_len'] = len(in_loading_futures)
+                stats['0_rowgroups_scheduled_for_loading'] += 1
```
rgruener: why prepend the 0

petastorm/reader_impl/worker_loop.py
```diff
@@ -49,15 +82,21 @@ def __str__(self):
...
-                stop_event, stats):
+                stop_event, row_serializer, stats):
     # Possibly infinite loop driven by row-group ventilation logic (infinite if epochs=inf)
     in_loading_futures = []
     in_decoding = []
 
     if not isinstance(stats, Counter):
         raise ValueError('stats argument is expected to be a collections.Counter')
 
+    chunk_decoding_size = max(1, decoding_pool._max_workers / _OUTPUT_QUEUE_SIZE)
```
rgruener: in python3 this can yield a float if `_max_workers` > 30. That seems like a bad thing. 

Also this seems to pretty much always be 1 since it should be rare to have more than 30 threads/processes anyways so Im not sure I understand the purpose of this.

petastorm/reader_impl/worker_loop.py
```diff
@@ -118,10 +160,10 @@ def worker_loop(epochs_generator, loading_pool, loader, decoding_pool, decoder,
...
             # ---------------------------------------------------------------------------------
             while shuffling_queue.can_retrieve() and len(in_decoding) < _MAX_IN_DECODING_QUEUE_SIZE:
-                shuffled = shuffling_queue.retrieve()
-                stats['rows_read_from_shuffling_queue'] += 1
-                in_decoding.append(decoding_pool.submit(_decode_row_dispatcher, decoder, shuffled))
-                stats['in_decoding_queue_len'] = len(in_decoding)
+                shuffled = shuffling_queue.retrieve_many(chunk_decoding_size)
+                stats['rows_read_from_shuffling_queue'] += len(shuffled)
+                from_shuffling.update(stats['rows_read_from_shuffling_queue'])
+                in_decoding.append(decoding_pool.submit(_decode_row_dispatcher, decoder, shuffled, row_serializer))
```
rgruener: you are passing `shuffled` (which can be a list) to the `_decode_row_dispatched` method which seems to still expect a single element.

petastorm/reader_impl/reader_v2.py
```diff
@@ -156,11 +155,13 @@ def __init__(self, dataset_url, schema_fields=None, shuffle=None, predicate=None
         epoch_items = self._apply_row_drop_partition(filtered_row_groups, shuffle_options)
 
         # 4. Launch a new thread running `worker_loop` function.
-        def epochs_iterator(): return epoch_generator(epoch_items, num_epochs, shuffle_options.shuffle_row_groups)
+        def epochs_iterator():
+            return epoch_generator(epoch_items, num_epochs, shuffle_options.shuffle_row_groups)
 
-        self._results_queue = Queue(_OUTPUT_QUEUE_SIZE)
+        self._results_queue = Queue()
```
rgruener: This removed the size of the results queue. It looks like it can now grow indefinitely?

petastorm/reader_impl/worker_loop.py
```diff
@@ -138,27 +180,38 @@ def worker_loop(epochs_generator, loading_pool, loader, decoding_pool, decoder,
...
+                            else:
+                                was_decoding_result = result_buffer
+
+                            output_queue.put(was_decoding_result, block=False)
+                            stats['rows_written_to_output_queue'] += len(was_decoding_result)
+                            from_decoding.update(stats['rows_written_to_output_queue'])
                             break
                         except queue.Full:
-                            sleep(_LOOP_IDLE_TIME_SEC)
+                            raise RuntimeError('Do not set Queue\'s maxsize. The size of the queue is managed by '
```
rgruener: I cant immediately see what is managing the size of the output queue. Is there something stopping it from growing indefinitely?

petastorm/reader_impl/worker_loop.py
selitvin: Yes. This does not make sense. Should build some sort of a benchmark to measure performance for different datasets structures to be smarter about this. Modifed `chunk_decoding_size` to become 1/10 of the rowgroup size when it becomes known.

petastorm/reader_impl/worker_loop.py
selitvin: It passes the parameters through to the row decoder. Change wrong names in `_decode_row_dispatcher` to reflect that.

petastorm/reader_impl/reader_v2.py
selitvin: The responsibility of the output queue control size is now worker_loop's. The code was missing, resurrected it.

petastorm/reader_impl/worker_loop.py
selitvin: Fixed. All the control mechanism of the worker_loop becomes tricky and probably is fragile. Will need to think how to improve it.

selitvin:
Please hold on with review. Found couple of serious logical bugs.

Issue #186, opened 2018-09-27
Unnecessary dependency on OpenCV


miguelvr:
I'm trying to export a non-image spark dataset to the petastorm format, and when I import the petastorm codecs, I get this:

```
~/.conda/envs/stb2/lib/python3.5/site-packages/petastorm/codecs.py in <module>()
     25 from io import BytesIO
     26 
---> 27 import cv2
     28 import numpy as np
     29 from pyspark.sql.types import BinaryType, LongType, IntegerType, ShortType, ByteType, StringType

ImportError: No module named 'cv2'
```

It doesn't make sense to keep opencv as a dependency, it seems unnecessary.

rgruener:
So we actually define opencv as an extra dependency in the setup.py however you are correct that it isnt really extra since we force it as a dependency (we mainly defined it as extra in case people want to install older versions as a debian package instead of through pip).

Since images are such a large part of what we use petastorm for we have not done anything to make it an optional dependency, though it certainly could be and probably should be.

PR #191, opened 2018-09-28
Fixes OpenCV requirement (#186)

miguelvr:
SImple fix for #186

### DIFF
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -20,14 +20,14 @@ transitively to parquet files.
 NOTE: Due to the way unischema is stored alongside dataset (with pickling), changing any of these codecs class names
 and fields can result in reader breakages.
 """
-import sys
 from abc import abstractmethod
 from io import BytesIO
 
 try:
     import cv2
+    OPENCV_AVAILABLE = True
 except ImportError:
-    pass
+    OPENCV_AVAILABLE = False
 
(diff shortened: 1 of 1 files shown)


petastorm/codecs.py
```diff
@@ -20,11 +20,15 @@
...
-
+import sys
 from abc import abstractmethod
 from io import BytesIO
 
-import cv2
+try:
+    import cv2
+except ImportError:
+    pass
```
rgruener: I think rather than using the assert below on `sys.modules` you can have a global variable `OPENCV_AVAILABLE` here to check whether the ImageCodec can be used

petastorm/codecs.py
```diff
@@ -53,6 +57,8 @@ def __init__(self, image_codec='png', quality=80):
         :param image_codec: any format string supported by opencv. e.g. ``png``, ``jpeg``
         :param quality: used when using ``jpeg`` lossy compression
         """
+        assert OPENCV_AVAILABLE, "{} requires opencv-python to be installed"
```
rgruener: Is `{}` in the string meant to be formatted?

miguelvr: yes, my bad. just fixed it

selitvin:
@miguelvr, thanks for creating the PR!

PR #197, opened 2018-10-03
ProcessPool output queue size control and zero copy pyarrow serialization

selitvin:
Effective control over the amount of rowgroups that are processed by process pool. The size of rowgroups allowed to be somewhere in loading/decoding is defined now by the amount of workers.

Improved the way we use pyarrow serialization and pyzmq to reduce amount of memory copies.

Improved pyarrow serialization to support decimal.Decimal type.

Improved diagnostics reported by Reader.

These are too many changes for a single PR, but they are split into small commits.

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -43,6 +43,10 @@ ShuffleOptions = petastorm.shuffle_options.ShuffleOptions
 # Make it easier to import as "from petastorm.reader import ReaderV2"
 ReaderV2 = ReaderV2
 
+# Ventilator guarantees that no more than workers + _VENTILATE_EXTRA_ROWGROUPS are processed at a moment by a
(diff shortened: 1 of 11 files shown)


petastorm/reader.py
```diff
@@ -43,6 +43,11 @@
 # Make it easier to import as "from petastorm.reader import ReaderV2"
 ReaderV2 = ReaderV2
 
+# Ventilator will hold ventilating new rowgroup if the current amount of 'rowgroups in processing' is greater than
```
rgruener: Not sure I understand the explanation of why we use this number of items in the ventilation queue.

rgruener:
I noticed this in my bundling PR but tests now seem to always hang on the process pool in python 3.6. Not really sure whats going on but I was not able to reproduce locally on my machine.

PR #198, opened 2018-10-04
Resurrect mnist examples and enabling unit tests

selitvin:
Trying to resolve previously observed segfaults by importing `pyarrow` before `torch` as explained here: https://github.com/selitvin/petastorm/blob/resurrect_mnist_examples/docs/troubleshoot.rst


examples/mnist/pytorch_example.py
```diff
@@ -116,7 +119,7 @@ def main():
                         help='learning rate (default: 0.01)')
     parser.add_argument('--momentum', type=float, default=0.5, metavar='M',
                         help='SGD momentum (default: 0.5)')
-    parser.add_argument('--no-cuda', action='store_true', default=False,
+    parser.add_argument('--no-cuda', action='store_true', default=True,
```
rgruener: for a store_true argument with default to True it is actually impossible to set it to false. Therefore either change the option to `use-cuda` or we would need 2 separate options for it (similar to how bops wors)

PR #205, opened 2018-10-08
Fixing ventilator race condition

selitvin:
If stop was called while ventilate_fn is in progress, the value of _iterations_remaining can become -1.
Introducing a separate flag to indicate forced end of ventilation.


petastorm/workers_pool/ventilator.py
```diff
@@ -117,7 +118,8 @@ def processed_item(self):
         self._processed_items_count += 1
 
     def completed(self):
-        return self._iterations_remaining == 0 or not self._items_to_ventilate
+        assert self._iterations_remaining >= 0
```
rgruener: `_iterations_remaining` can be None to specify unlimited epochs

PR #211, opened 2018-10-22
Make petastorm.pytorch.Dataloader compatible with additional integer types.

selitvin:
PyTorch supports only following integer types: int8, int32 and int64. PyTorch does not
support tensors of strings/objects. Adding an extra stage in `petastorm.pytorch.Dataloader`
that promotes unsupported integer types to the next (wider) integer type that is supported.

Raising an exception with a clear message if a value is unsupported and can not be
promoted, such as `np.string_`, `np.unicode_` or `np.object`.

### DIFF
--- a/petastorm/pytorch.py
+++ b/petastorm/pytorch.py
@@ -15,7 +15,65 @@
 # Must import pyarrow before torch. See: https://github.com/uber/petastorm/blob/master/docs/troubleshoot.rst
 import pyarrow  # noqa: F401 pylint: disable=W0611
 
+import collections
+import decimal
+
+import numpy as np
+from six import PY2
 from torch.utils.data.dataloader import default_collate
+
+if PY2:
+    _string_classes = basestring  # noqa: F821
+else:
+    _string_classes = (str, bytes)
+
+
+def _sanitize_pytorch_types(row_as_dict):
+    """Promotes values types in a dictionary to the types supported by pytorch. Raises an error if type is clear error
+    if the type can not be promoted.
+
+    The parameter is modified in-place.
+
+    int8, uint16 are promoted to int32; uint32 -> int64;
+    numpy string_, unicode_, object arrays are not supported.
+
+    :param row_as_dict: a dictinoary of key-value pairs. The values types are promoted to pytorch compatible.
+    :return: None
+    """
+    for name, value in row_as_dict.items():
+        # PyTorch supported types are: double, float, float16, int64, int32, and uint8
+        if isinstance(value, np.ndarray):
+            if value.dtype == np.int8:
+                row_as_dict[name] = value.astype(np.int16)
+            elif value.dtype == np.uint16:
+                row_as_dict[name] = value.astype(np.int32)
+            elif value.dtype == np.uint32:
+                row_as_dict[name] = value.astype(np.int64)
+
+
+def decimal_friendly_collate(batch):
+    """A wrapper on top of ``default_collate`` function that allows decimal.Decimal types to be collated.
+
+    We use ``decimal.Decimal`` types in petastorm dataset to represent timestamps. PyTorch's ``default_collate``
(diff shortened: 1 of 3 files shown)


petastorm/pytorch.py
```diff
@@ -13,11 +13,74 @@
...
+    numpy string_, unicode_, object arrays are not supported.
+
+    :param row_as_dict: a dictinoary of key-value pairs. The values types are promoted to pytorch compatible.
+    :return: None
+    """
+    for name, value in row_as_dict.items():
+        # PyTorch supported types are: double, float, float16, int64, int32, and uint8
+        if isinstance(value, np.ndarray):
+            if value.dtype == np.int8:
+                row_as_dict[name] = value.astype(np.int16)
```
davifrossard: `int8` (and `uint8`) have native pytorch types, so they dont need upcasting

petastorm/pytorch.py
```diff
@@ -58,7 +121,9 @@ def __iter__(self):
             # Default collate does not work nicely on namedtuples and treat them as lists
             # Using dict will result in the yielded structures being dicts as well
             row_as_dict = row._asdict()
-            batch.append(self.transform(row_as_dict) if self.transform else row_as_dict)
+            _sanitize_pytorch_types(row_as_dict)
+            transformed_row = self.transform(row_as_dict) if self.transform else row_as_dict
```
davifrossard: The transforms should be called first, no? (otherwise the user doesnt have a chance to convert strings to something else before the `NotImplementedError` is raised in `_sanitize_pytorch_types`)

petastorm/pytorch.py
```diff
@@ -13,11 +13,74 @@
...
+    :param batch: A list of dictionaries to collate
+    :return: A dictionary of lists/pytorch.Tensor types
+    """
+
+    if isinstance(batch[0], decimal.Decimal):
+        return batch
+    elif isinstance(batch[0], collections.Mapping):
+        return {key: decimal_friendly_collate([d[key] for d in batch]) for key in batch[0]}
+    elif isinstance(batch[0], _string_classes):
+        return batch
```
davifrossard: Would this case ever occur? (since `_sanitize_pytorch_types` blocks strings)

petastorm/pytorch.py
selitvin: This is what I see (`int8` appears to be unsupported, while `uint8` is supported):
```
In [1]: import numpy as np; import torch

In [2]: torch.as_tensor(np.asarray([1,2]).astype(np.int8))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-02b070359a1b> in <module>()
----> 1 torch.as_tensor(np.asarray([1,2]).astype(np.int8))

TypeError: can't convert np.ndarray of type numpy.int8. The only supported types are: double, float, float16, int64, int32, and uint8.

In [3]: torch.as_tensor(np.asarray([1,2]).astype(np.uint8))
Out[3]: tensor([1, 2], dtype=torch.uint8)
```

petastorm/pytorch.py
selitvin: Will not raise an error in case of strings/objects in `_sanitize_pytorch_types`. Will leave a chance for a user to update these types in transform. Added a test case that covers that scenario. Does it make sense?

petastorm/pytorch.py
selitvin: Removed the strings/object blockage from `_sanitize_pytorch_types`. Even before that, string scalars were not triggering an error, only numpy arrays of strings did.

petastorm/pytorch.py
davifrossard: You're right. Seems to be a mistake in how pytorch handles the type, considering you can create tensors of type `int8`, in any case the casting is safe so I guess it doesnt really matter.

```
In [23]: torch.CharTensor(np.zeros(1, dtype=np.int16))
Out[23]: tensor([0], dtype=torch.int8)
```

PR #216, opened 2018-10-24
Transmit _WORKER_STARTED_INDICATOR when process pool worker is ready

selitvin:
Previously was relying on monitoring socket. Sometimes, a message was lost resulting in indefinite wait for all
workers to start. The new method is simpler and appears to be more robust.

### DIFF
--- a/petastorm/workers_pool/process_pool.py
+++ b/petastorm/workers_pool/process_pool.py
@@ -23,7 +23,6 @@ from traceback import format_exc
 
 import zmq
 from zmq import ZMQBaseError
-from zmq.utils import monitor
 
 from petastorm.reader_impl.pickle_serializer import PickleSerializer
 from petastorm.reader_impl.pyarrow_serializer import PyArrowSerializer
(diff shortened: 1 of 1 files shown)


petastorm/workers_pool/process_pool.py
```diff
@@ -42,9 +41,36 @@
...
+#
+# ----------------                                    ------------------
+# |              |  --- _ventilator_send  (push) -->  |                |
+# | main process |  --- _control_sender   (pub)  -->  | worker process |
+# |              |  <-- _results_receiver (pull)  --  |                |
+# ----------------                                    ------------------
+#
+# 1. When ProcessPool start is called, it creates _ventilator_send, _control_sender and _result_receiver
+#    sockets.
+# 2. After initialization is done, worker process sends _WORKER_STARTED_INDICATOR
```
rgruener: If the _WORKER_STARTED_INDICATOR is sent by the worker before the results_receiver socket is setup is it possible to lose the indicator causing blocking/timeouts. Im not sure if thats a possibility but its something I vaguely remember seeing. I think its something we can keep in mind if we start seeing failures for that.

PR #232, opened 2018-10-30
Add ability to split row groups by reading parquet footers

rgruener:


petastorm/etl/dataset_metadata.py
```diff
@@ -298,6 +306,26 @@ def _split_row_groups(dataset):
...
+
+    def split_piece(piece):
+        metadata = piece.get_metadata(dataset.fs.open)
+        return [pq.ParquetDatasetPiece(piece.path,
+                                       row_group,
+                                       piece.partition_keys) for row_group in range(metadata.num_row_groups)]
+
+    futures_list = [thread_pool.submit(split_piece, piece) for piece in dataset.pieces]
+
+    if futures_list:
```
selitvin: Is there a reason for an explicit wait? Is it always ok to access futures after the pool is shutdown (even with the wait)? Would expect something simpler here:
```python
result =  [item for f in futures_list for item in f.result()]
thread_pool.shutdown()
return result
```

petastorm/reader.py
```diff
@@ -101,7 +101,8 @@ def make_reader(dataset_url,
     :param hdfs_driver: A string denoting the hdfs driver to use (if using a dataset on hdfs). Current choices are
         libhdfs (java through JNI) or libhdfs3 (C++)
     :param infer_schema: Whether to infer the unischema object from the parquet schema.
-            Only works for schemas containing certain scalar type.
+            Only works for schemas containing certain scalar type. This option allows getting around using
```
selitvin: "around using" maybe something more explicit, like "around explicitly generating petastorm metadata using petastorm-generate-metadata.py"?

petastorm/etl/dataset_metadata.py
rgruener: Ah true, I was thinking that we need to wait for all the futures to complete (which is true) though the explicit wait is indeed not necessary

rgruener:
> Are you sure setting infer_schema=False does not create unnecessary barrier for a user? Isn't the worst scenario if the metadata is not there is that the user would get a message with instructions on how to fix?

So I think until we fix the performance of these out of the box parquet datasets I am ok with it being a bit of a barrier to use them. We could simply look for the petastorm specific metadata and fall back to automatically infer things however I worry that for a random user they would be turned off by the poor performance we usually have with small rows of scalars. It might be better to have a bit of a barrier to have users understand this is a work in progress and not a highly supported usecase right now.

PR #236, opened 2018-11-11
Adding batching_table_queue.py module.

selitvin:
Include BatchingTableQueue class acts as a FIFO queue, taking rows in the form of Arrow Table object returning
batches of rows of a fixed length, also as Arrow Table objects.

This is one of the utilities we need to work with Arrow Tables and support batches as reader result.


petastorm/pyarrow_helpers/batching_table_queue.py
```diff
@@ -0,0 +1,79 @@
...
+        record_batches = table.to_batches()
+        for record_batch in record_batches:
+            self._buffer.append(record_batch)
+            self._cumulative_len += record_batch.num_rows
+
+    def empty(self):
+        """Checks if more tables can be returned by get. If the number of rows in the internal buffer is less then
+        ``batch_size``, empty would return False.
+        """
+        return self._head_idx + self._batch_size > self._cumulative_len
```
rgruener: im curious what happens for the last batch? Once you are at the end of an epoch you might not have enough to fill the final batch completely which would result in throwing out those final few rows?

selitvin: Yes. I chose not to implement that behavior for now as it would require signaling that we are at the end of input, but we might want to extend in the future.

PR #251, opened 2018-12-02
Support native parquet stores with scalar-only Parquet types

selitvin:
Adding `make_parquet_reader`: support for working efficiently with parquet stores.
    
Currently only stores with scalars are supported.
Readers created with `make_parquet_reader` return batches of rows and not row-by-row, as `make_reader`.


petastorm/tests/test_parquet_reader.py
```diff
@@ -0,0 +1,57 @@
...
+    assert count == len(expected_data)
+
+
+@pytest.mark.parametrize('reader_factory', _D + _TP)
+def test_simple_read(scalar_dataset, reader_factory):
+    """Just a bunch of read and compares of all values to the expected values using the different reader pools"""
+    with reader_factory(scalar_dataset.url) as reader:
+        _check_simple_reader(reader, scalar_dataset.data)
+
+# TODO(yevgeni): missing tests: subset of columns; invalid column name;
```
rgruener: Maybe attach a github issue to this todo?

petastorm/tests/test_end_to_end_predicates_impl.py
```diff
@@ -37,3 +37,21 @@ def get_fields(self):
...
+
+class VectorizedEqualPredicate(PredicateBase):
+    def __init__(self, values):
+        self._values = values
+
+    def get_fields(self):
+        return list(self._values.keys())
+
+    def do_include(self, values):
+        result = None
```
rgruener: It might be more concise/clear to write this as:
```
        result = [True] * len(values)
        for field_name in self._values.keys():
            result &= values[field_name] == self._values[field_name]
        return result
```

petastorm/tests/test_end_to_end.py
```diff
@@ -89,24 +119,24 @@ def test_simple_read_with_disk_cache(synthetic_dataset, reader_factory, tmpdir):
     with reader_factory(synthetic_dataset.url, num_epochs=2,
                         cache_type='local-disk', cache_location=tmpdir.strpath,
                         cache_size_limit=CACHE_SIZE, cache_row_size_estimate=ROW_SIZE_BYTES) as reader:
-        _check_simple_reader(reader, synthetic_dataset.data, 200)
+        list(reader)
```
rgruener: this no longer checks the ids being read, I understand that this test is more to check that we can read with the disk cache without failures but is there a reason not to check that the ids make sense as well?

petastorm/reader.py
```diff
@@ -241,6 +346,18 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
...
             By default, use the :class:`.NullCache` implementation.
+
+        :param infer_schema: Whether to infer the unischema object from the parquet schema.
+            Only works for schemas containing certain scalar type. This option allows getting around explicitly
+            generating petastorm metadata using :func:`petastorm.etl.dataset_metadata.materialize_dataset` or
+            petastorm-generate-metadata.py. Should always be True for non-Petastorm generated datasets.
+
+        :param worker_class: This is the class that will be instantiated on a different thread/process. It's
+            responsibility is to load and filter the data.
+        :param results_queue_reader: An instance of an object that will be reading data produced by worker class.
```
rgruener: Since this is tightly coupled with the worker class does it make sense to have the results_queue_reader as a separate parameter?

petastorm/tests/test_end_to_end_predicates_impl.py
rgruener: Though actually that might just be my functional programming preference actually

petastorm/tests/test_end_to_end.py
selitvin: I removed it because _check_simple_reader does not support batches. Replaced it with the test on ids. That's concise enough.

petastorm/reader.py
selitvin: Fixed.

selitvin:
Renamed `make_parquet_reader` -> `make_batch_reader`
Added explicit checks to these functions to forbid opening Petastorm stores from `make_batch_reader` and native Parquet stores using `make_reader`.

petastorm/tests/test_end_to_end.py
selitvin: Added an id check. Checking all fields is a bit trickier (though possible). Do you think it's needed?

rgruener: I dont think its necessary. I think that this test doesnt need to cover that logic.

petastorm/unischema.py
```diff
@@ -206,7 +208,13 @@ def from_arrow_schema(cls, arrow_schema):
         :param arrow_schema: :class:`pyarrow.lib.Schema`
         :return: A :class:`Unischema` object.
         """
+        meta = parquet_dataset.pieces[0].get_metadata(parquet_dataset.fs.open)
+        arrow_schema = meta.schema.to_arrow_schema()
         unischema_fields = []
+
+        for partition_name in parquet_dataset.partitions.partition_names:
+            unischema_fields.append(UnischemaField(partition_name, np.str_, (), ScalarCodec(StringType()), False))
```
rgruener: nice catch on the partition columns!

PR #269, opened 2018-12-19
Make reading a non-petastorm dataset fail if contains Null values

selitvin:
Previously, unexpected conversion to float was occuring when an int column contained Nulls


selitvin:
Planning changes

PR #278, opened 2018-12-29
Adding transform-on-worker thread/pool functionality.

selitvin:
A user needs to specify transform_spec when calling `make_reader`.
The transform-spec defines both the function and the post-transform scheme of the data.

output_schema attribute was added to the Reader interface.

TODO list (will open issues once the PR lands)
- Add transform examples to the tutorials
- Add proper unit test for the new transform.py module
- Add transform functionality to `make_batch_reader`
- Validate the return value of the transform function conforms to the post-transform schema and raise an error if not.
- Verify transform functionality when using ngram

### DIFF
--- a/petastorm/__init__.py
+++ b/petastorm/__init__.py
@@ -13,5 +13,6 @@
 # limitations under the License.
 
 from petastorm.reader import make_reader, make_batch_reader  # noqa: F401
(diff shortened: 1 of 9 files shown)


selitvin:
All changes have landed as part of #292

Issue #287, opened 2019-01-16
pyarrow.hdfs.HadoopFileSystem not serializable by Spark


Limmen:
https://github.com/uber/petastorm/blob/e65e327f3bf10ffe95954ceed1a5da89ea0b0ba2/petastorm/etl/dataset_metadata.py#L212

Hi again,

I am using pyspark 2.4.0, and pyarrow 0.11.1 and spark is not able to serialize pyarrow.hdfs.HadoopFileSystem. Have you encountered this issue before? 

I get "HDFS Connection Failed" during the serialization, which is a bit strange. I can open a new filesystem connection inside the spark-mapper with no problem, so there should not be a connection issue but rather a serialization issue. 

Snippet from stacktrace:
`
  File "/srv/hops/hopsdata/tmp/nm-local-dir/usercache/N8YmHUGK9tr5Q9_iz7ZdAb0oU66QXgWDdYzH4tE4wgI/appcache/application_1547648243443_0001/container_e01_1547648243443_0001_01_000002/pyspark.zip/pyspark/serializers.py", line 566, in loads
    return pickle.loads(obj, encoding=encoding)
  File "/srv/hops/anaconda/anaconda/envs/petastorm/lib/python3.6/site-packages/pyarrow/hdfs.py", line 37, in __init__
    self._connect(host, port, user, kerb_ticket, driver, extra_conf)
  File "pyarrow/io-hdfs.pxi", line 105, in pyarrow.lib.HadoopFileSystem._connect
  File "pyarrow/error.pxi", line 83, in pyarrow.lib.check_status
pyarrow.lib.ArrowIOError: HDFS connection failed
`

Thanks
/Kim

selitvin:
I think it is expected for a resource connection not to be serializable. For a file system connection, as in this case, it may include file handles which are not valid outside of the process that `open`ed them.

When we distribute large evaluation task using spark we create Petastorm reader in the function invoked by `map`. From your question, I understand that this solution works for you as well. Would you be able to follow this pattern in your case?

Limmen:
Yes exactly. I can get it working in my fork of petastorm by opening the filesystem connection inside the mapper. However I can not get it working using the upstream version of petastorm. Inside `dataset_metadata.py` there is a spark-mapper that requires the filesystem object to be serialized (function `get_row_group_info(path)` in file `dataset_metadata.py`)

When I try to use `with materialize_dataset(...)` using libhdfs on our cluster I get the serialization issue described above.

[...]

PR #289, opened 2019-01-17
Parameterize the hdfs-driver in the API to support libhdfs

Limmen:
In methods such as `materialize_dataset` the user can supply its own fs-handle
to use libhdfs instead of libhdfs3. But in other API methods, such as `dataset_as_rdd`,
there is currently no way for the user to specify to use libhdfs over libhdfs3.
This commit fixes that.

For example, before this commit, running the following command in a libhdfs environment:
```python
dataset_as_rdd(dataset_url, spark, [HelloWorldSchema.id, HelloWorldSchema.image1])
```
would give:
```
pyarrow.lib.ArrowIOError: Unable to load libhdfs3
```
With the changes of this commit you can specify the optional argument `hdfs_driver`
```
rdd = dataset_as_rdd(dataset_url, spark, [HelloWorldSchema.id, HelloWorldSchema.image1], hdfs_driver="libhdfs")
```
so that you can overwrite default value `libhdfs3` and make it compatible with libhdfs environments as well.

This PR is mostly to demonstrate what the minimal changes needed to make petastorm working with libhdfs on our cluster. Using these fixes in addition to a workaround for this bug: https://github.com/uber/petastorm/issues/287 (this is not included in this commit as it requires a bigger change to the API, we just have a temporary workaround in our fork at the moment, waiting for the next release of petastorm) all petastorm examples are working on our cluster with libhdfs.

### DIFF
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -365,13 +365,15 @@ def get_schema(dataset):
     return schema
 
 
-def get_schema_from_dataset_url(dataset_url):
+def get_schema_from_dataset_url(dataset_url, hdfs_driver='libhdfs3'):
     """Returns a :class:`petastorm.unischema.Unischema` object loaded from a dataset specified by a url.
 
     :param dataset_url: A dataset URL
+    :param hdfs_driver: A string denoting the hdfs driver to use (if using a dataset on hdfs). Current choices are
+        libhdfs (java through JNI) or libhdfs3 (C++)
     :return: A :class:`petastorm.unischema.Unischema` object
     """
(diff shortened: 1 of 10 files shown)


selitvin:
@Limmen , thank you for the contribution! The build has failed for unrelated reason (I saw it hangs like this in the past). Until we fix the build (#290), restarting the build manually.

selitvin:
Looks good to me. I can land it if you are done with the PR.

Limmen:
@selitvin Cool, the PR is done from my side as well.

selitvin:
@Limmen , thank you for that! Landing....

PR #292, opened 2019-01-18
transform= argument in make_batch_reader

selitvin:


petastorm/reader.py
```diff
@@ -411,6 +426,10 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
         # Will raise an exception if invalid schema fields are in schema_fields
         fields = schema_fields if isinstance(schema_fields, collections.Iterable) else None
         self.schema = stored_schema.create_schema_view(fields) if fields else stored_schema
+        if transform_spec:
+            self.output_schema = transform_schema(self.schema, transform_spec)
```
rgruener: Using `self.output_schema` is kind of an api breaking change since that now represents the schema of rows that the reader will give you. I think it might be better to have `reader.schema` remain as the schema that the reader returns and then perhaps have a `reader.storage_schema` be the schema that the underlying storage has.

To me this makes sense since I would expect that users care more about the schema of rows the reader returns rather than the schema of the underlying storage.

selitvin: Thank you! Indeed much cleaner that way. Fixed this in #278 and rebased this PR on top.

Issue #286, opened 2019-01-15
Create an 'hello-world' example for non-Petastorm dataset usage


selitvin:
We introduced `make_batch_reader` that allows user to access Parquet stores that were not created with Petastorm (i.e. without Petastorm metadata). There is no concise example showing that.

jsgoller1:
Resolved by #293

PR #293, opened 2019-01-18
Adds example for creating non-petastorm data sets and reading them with petastorm

jsgoller1:
**Problem**
The Petastorm docs claim that it is possible to access Parquet data stores that were not created with Petastorm. There is no documentation denoting how to do this, so we need to add it. (#286)

**Solution**
- [x] Read it in via Petastorm with the batch reader.
- [x] Create a parquet data store through raw `pyspark`

**Additional**
- [x] Fix linter errors
- [x] Separate `hello_world` into two directories, with one for non-Petastorm data examples
- [x] Replicate other existing examples in non-Petastorm data directory
- [x] Add test for non-Petastorm example


examples/hello_world/python_non_petastorm_hello_world.py
```diff
@@ -0,0 +1,32 @@
...
+
+from __future__ import print_function
+
+from petastorm import make_batch_reader
+
+
+def python_non_petastorm_hello_world(dataset_url='file:///tmp/non_petastorm_dataset'):
+    # Reading data from the non-Petastorm Parquet file with inferred Unischema
+    with make_batch_reader(dataset_url, schema_fields=["id", "value1", "value2", "value3", "value4"]) as reader:
+        schema_view = reader.next()
```
jsgoller1: @selitvin Is there a `correct` way to get column data from `Reader` other than by calling `.next()`? This feels wrong.

rgruener: A `Reader` is essentially a python iterator so the correct way to get data from it would be to iterate through it or use `next(reader)` for a single row

examples/hello_world/python_non_petastorm_hello_world.py
```diff
@@ -0,0 +1,32 @@
...
+"""Minimal example of how to read samples from a dataset generated by `generate_non_petastorm_dataset.py`
+using plain Python"""
+
+from __future__ import print_function
+
+from petastorm import make_batch_reader
+
+
+def python_non_petastorm_hello_world(dataset_url='file:///tmp/non_petastorm_dataset'):
+    # Reading data from the non-Petastorm Parquet file with inferred Unischema
```
selitvin: Maybe we can even not mention Unischema here, it's just Petastorm reading a standard Parquet store. I think Unischema added value stems from the support of tensor types. For a plain Parquet store it is not relevant.

examples/hello_world/python_non_petastorm_hello_world.py
```diff
@@ -0,0 +1,32 @@
...
+    with make_batch_reader(dataset_url, schema_fields=["id", "value1", "value2", "value3", "value4"]) as reader:
+        schema_view = reader.next()
+        for i, val1, val2 in zip(schema_view.id, schema_view.value1, schema_view.value2):
```
selitvin: Maybe worth mentioning in a comment that the return values are batches of records?

examples/hello_world/external_dataset/tensorflow_hello_world.py
```diff
@@ -0,0 +1,45 @@
...
+from petastorm.tf_utils import tf_tensors, make_petastorm_dataset
+
+
+def tensorflow_hello_world(dataset_url='file:///tmp/external_dataset'):
+    # Example: tf_tensors will return tensors with dataset data
+    with make_batch_reader(dataset_url) as reader:
+        tensor = tf_tensors(reader)
+        with tf.Session() as sess:
+            sample = sess.run(tensor)
+            print(sample.id)
```
jsgoller1: @selitvin this winds up printing a group of `id`s instead of just one due to the batch reader - should we change it so that each id is printed individually? Otherwise, I can just add a comment.

examples/hello_world/external_dataset/pytorch_hello_world.py
```diff
@@ -0,0 +1,31 @@
...
+from __future__ import print_function
+
+from petastorm import make_batch_reader
+from petastorm.pytorch import DataLoader
+
+
+def pytorch_hello_world(dataset_url='file:///tmp/external_dataset'):
+    with DataLoader(make_batch_reader(dataset_url)) as train_loader:
+        sample = next(iter(train_loader))
+        print(sample['id'])
```
jsgoller1: @selitvin this prints a group of `id`s too, so the same question applies.

examples/hello_world/external_dataset/generate_external_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+        .master('local[2]')\
+        .getOrCreate()
+    sc = spark.sparkContext
+
+    rows_count = 10
+    rows_rdd = sc.parallelize(range(rows_count))\
+        .map(row_generator)
+
+    spark.createDataFrame(rows_rdd, NON_PETASTORM_SCHEMA).\
+        coalesce(rows_count).\
```
selitvin: We probably don't need the coalesce here.

examples/hello_world/external_dataset/generate_external_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+def row_generator(x):
+    """Returns a single entry in the generated dataset. Return a bunch of random values as an example."""
+    return {'id': x,
+            'value1': random.randint(-255, 255),
+            'value2': random.randint(-255, 255)}
+
+
+def generate_external_dataset(output_url='file:///tmp/external_dataset'):
+    """Creates an example dataset at output_url in Parquet format"""
+    spark = SparkSession.builder.config('spark.driver.memory', '2g')\
```
selitvin: Can you please try without the `spark.driver.memory`? We probably don't need it and it would remove unneeded bloat.

examples/hello_world/external_dataset/generate_external_dataset.py
```diff
@@ -0,0 +1,58 @@
...
+        .master('local[2]')\
+        .getOrCreate()
+    sc = spark.sparkContext
+
+    rows_count = 10
+    rows_rdd = sc.parallelize(range(rows_count))\
+        .map(row_generator)
+
+    spark.createDataFrame(rows_rdd, NON_PETASTORM_SCHEMA).\
```
selitvin: We don't have to specify `NON_PETATSTORM_SCHEMA` in the call at all (and remove the variable). Spark will guess the schema from the data. That would make the example even simpler.

examples/hello_world/external_dataset/python_hello_world.py
```diff
@@ -0,0 +1,33 @@
...
+
+from petastorm import make_batch_reader
+
+
+def python_hello_world(dataset_url='file:///tmp/external_dataset'):
+    # Reading data from the non-Petastorm Parquet via pure Python
+    with make_batch_reader(dataset_url, schema_fields=["id", "value1", "value2"]) as reader:
+        for schema_view in reader:
+            # Batches are returned
+            for i, val1, val2 in zip(schema_view.id, schema_view.value1, schema_view.value2):
```
selitvin: I think it is ok to show that batches are coming out of the reader. That way all samples do the same.

examples/hello_world/external_dataset/tensorflow_hello_world.py
selitvin: Sure. Feel free to do whatever you feel makes it easiest for the new user to understand what we are doing here. Maybe a comment, you can also add a print statement that explains the output.

examples/hello_world/external_dataset/generate_external_dataset.py
jsgoller1: Removed.

examples/hello_world/external_dataset/generate_external_dataset.py
jsgoller1: Removing the schema throws a warning about inferring schema from an RDD of dicts being deprecated, so I've modified it to use Rows instead.

Issue #168, opened 2018-09-20
Support Petastorm on Macos


selitvin:
Make sure petastorm is installable on macos
Add macos as an additional OS on travis

PR #295, opened 2019-01-21
Create Dockerized workspace

jsgoller1:
# Problem
Per #168, Petastorm is currently not supported on MacOS, which makes development tricky for Mac users (such as myself). Although a deeper exploration of this issue is warranted, a quick fix for now is to create a platform-independent workspace via Docker. 

The workspace container should have all necessary Petastorm dependencies installed, but not have the source added - we can instead mount the local filesystem in with `-v`, which will free us from having to rebuild the container each time we make changes to the source. I have opted to delete the portions of the Dockerfile that add the source since the most recent commit is volume-mounted in the `.travis` file, and it only takes up extra space during push/pull.

# Solution
- [x] Reorganize and clean up Dockerfile, and remove portions that add local Petastorm source during build
- [x] Add Makefile to automate the above, as well as other operational tasks
- [x] Handle Travis related changes; move CI container build process to Makefile, and make slight tweak to `.travis` file.
- [x] Run through all Makefile commands and the `hello_world` examples to ensure workspace is functional.


docker/base.dockerfile
```diff
@@ -1,51 +1,46 @@
...
 ADD setup.py /petastorm/
+ADD README.rst /petastorm/
 ADD petastorm /petastorm/petastorm
 
+# Set up Python3 environment
+RUN python3.6 -m pip install pip --upgrade
+RUN python3.6 -m pip install wheel
 RUN python3.6 -m venv /petastorm_venv3.6
 RUN /petastorm_venv3.6/bin/pip3.6 install -e /petastorm/[test,tf,torch,docs,opencv]
 RUN /petastorm_venv3.6/bin/pip3.6 uninstall -y petastorm
```
jsgoller1: As far as I can tell, this line (and the repeated one in the Python3.6 section) don't actually do anything.

docker/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+ADD setup.py /petastorm/
+ADD README.rst /petastorm/
+ADD petastorm /petastorm/petastorm
+
+# Set up Python3 environment
+RUN python3.6 -m pip install pip --upgrade
+RUN python3.6 -m pip install wheel
+RUN python3.6 -m venv /petastorm_venv3.6
+RUN /petastorm_venv3.6/bin/pip3.6 install -e /petastorm/[test,tf,torch,docs,opencv]
+RUN /petastorm_venv3.6/bin/pip3.6 uninstall -y petastorm
```
jsgoller1: I am not sure if this line (or the following one in the python2.7 section) do anything, as Petastorm isn't being installed during this process.

jsgoller1:
@rgruener will do.

docker/Dockerfile
selitvin: `pip3.6 install -e /petastorm/[test,tf,torch,docs,opencv]` does install petastorm into the virtual environment in editable mode. I did not see a way to easily install petastorm dependencies without installing petastorm, so my line of thought was to install petastorm + dependencies and then remove petastorm, hence ending up just with the dependencies.

jsgoller1:
@rgruener Added docs, will merge once the Travis build finishes.

PR #299, opened 2019-01-23
Add stats collector for basic reader metrics

rgruener:
This is just the most basic metrics (which are just within the reader class only). My plan is to have a separate pr for incorporating worker pool metrics (e.g. results queue size) and merging with the diagnostics

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -36,6 +36,7 @@ from petastorm.reader_impl.reader_v2 import ReaderV2
 from petastorm.reader_impl.same_thread_executor import SameThreadExecutor
 from petastorm.reader_impl.shuffling_buffer import NoopShufflingBuffer, RandomShufflingBuffer
 from petastorm.selectors import RowGroupSelectorBase
+from petastorm.stats import NoopStatisticsCollector
 from petastorm.transform import transform_schema
 from petastorm.workers_pool.dummy_pool import DummyPool
 from petastorm.workers_pool.process_pool import ProcessPool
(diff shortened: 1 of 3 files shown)


petastorm/reader.py
```diff
@@ -603,11 +620,14 @@ def __iter__(self):
...
     def __next__(self):
-        try:
-            return self._results_queue_reader.read_next(self._workers_pool, self.schema, self.ngram)
-        except StopIteration:
-            self.last_row_consumed = True
-            raise
+        with self._statistics_collector.time_row_retrieval():
+            try:
+                results = self._results_queue_reader.read_next(self._workers_pool, self.schema, self.ngram)
+                self._statistics_collector.count_rows_retrieved()
```
selitvin: The name is a little misleading for the batch case, as it would count the batches. 

Would it make sense to blend `count_rows_retrieved` with `time_row_retrieval`? For example Petastorm could provide a 'performance counter' implementation (e.g. aggregate, number of measures, average/last/median an event took, etc). A customization piece should be just how the reporting is done.

rgruener: Yeah I was thinking about the name in the batch case.

And I think that is something we can provide

rgruener: So median is difficult to do since in order to calculate the median you need to store all values seen (could run into memory issues potentially). We could calculate means but not sure how much value that would add.

Also im trying to think of better names and nothing is obvious to me other than simply calling it 'next calls' or something like that.

Issue #280, opened 2019-01-02
A string type is always assumed when filtering by a field used as a partition


GregAru:
I am trying to read data from a partitioned dataset in S3 where the partition column is defined as an integer:
`UnischemaField('partcol', np.int32, (), ScalarCodec(IntegerType()), False),`

The code below executes without exceptions but returns nothing:
```   
with make_reader(dataset_path, 
                     workers_count=1,
                     predicate=in_lambda(['partcol'], lambda partcol: (partcol == 20190102))
                     ) as reader:
```

The following code returns the expected data from the partition:
```
 with make_reader(dataset_path,
                     workers_count=1,
                     predicate=in_lambda(['partcol'], lambda partcol: (partcol == '20190102'))
                     ) as reader:
```

selitvin:
Yep, we can address this since we know the type from Unischema.

jsgoller1:
@GregAru - looking into this.

jsgoller1:
@GregAru - while trying to reproduce the issue locally, I'm actually seeing the _opposite_ (though still incorrect) behavior, where a string passed as the partition column causes no data to be read, but an integer does:

```
    with make_reader(PETASTORM_DATASET_PATH,
                     workers_count=1,
                     predicate=in_lambda(['partcol'], lambda partcol: (partcol == '2'))
                     ) as reader:
        for schema_view in reader:
            print("This should return a schema_view:")
            print(schema_view)

    with make_reader(PETASTORM_DATASET_PATH,
                     workers_count=1,
                     predicate=in_lambda(['partcol'], lambda partcol: (partcol == 2))
                     ) as reader:
        for schema_view in reader:
            print("This should never be reached.")
            print(schema_view)
```
outputs the following:
```
(petastorm_venv3.6) root@computer:/petastorm# python test.py
This should never be reached.
HelloWorldSchema_view(id=2, partcol=2)
```
Is this at all consistent with what you are seeing (i.e. is the reported issue backwards)?

GregAru:
@jsgoller1 - Not quite. In my case, the following does not return data:
```
     with make_reader(PETASTORM_DATASET_PATH,
                     workers_count=1,
                     predicate=in_lambda(['partcol'], lambda partcol: (partcol == 2))
                     ) as reader:
        for schema_view in reader:
            print("This should never be reached.")
            print(schema_view)
```

Please note that I am reading the dataset from S3 and the dataset is partitioned by `partcol `

jsgoller1:
[...]

PR #307, opened 2019-02-06
Interpret partition values as correct type from schema when applying predicate

jsgoller1:
## Issue
#280 

## Problem
Currently, if a `Reader` is initialized on a partitioned dataset with a predicate, the partition values will be interpreted as strings during predicate application, which is problematic in cases like `"1234"` where the partition value should be read as an int.

## Solution
- [X] Correctly infer type by obtaining it from the unischema
- [x] Add unit test that generates a partitioned dataset and confirms functionality.

### DIFF
--- a/petastorm/tests/test_predicates.py
+++ b/petastorm/tests/test_predicates.py
@@ -14,10 +14,17 @@
 
 import pytest
 
+import numpy as np
+from pyspark.sql import SparkSession
+from pyspark.sql.types import IntegerType
+
 from petastorm import make_reader
+from petastorm.codecs import ScalarCodec
+from petastorm.etl.dataset_metadata import materialize_dataset
 from petastorm.predicates import in_set, in_intersection, \
     in_negate, in_reduce, in_pseudorandom_split, in_lambda
 from petastorm.tests.test_common import TestSchema
+from petastorm.unischema import dict_to_spark_row, Unischema, UnischemaField
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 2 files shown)


examples/hello_world/petastorm_dataset/generate_petastorm_dataset.py
```diff
@@ -40,7 +40,7 @@ def row_generator(x):
             'array_4d': np.random.randint(0, 255, dtype=np.uint8, size=(4, 128, 30, 3))}
 
 
-def generate_petastorm_dataset(output_url='file:///tmp/hello_world_dataset'):
+def generate_petastorm_dataset(output_url='file:///tmp/hello_world_dataset', partitioned=False):
```
selitvin: Would suggest to keep the hello world example as minimal as possible. Partitioning is not a trivial feature and may confuse. Would suggest to copy this code into the test file itself.

Also, having a test depend on an example code is unexpected (unless the test is checking the example)

jsgoller1: Got it; I was attempting to minimize code replication - I'll undo this change and copy it into the test case.

petastorm/tests/test_predicates.py
```diff
@@ -124,3 +131,51 @@ def test_predicate_on_single_column(synthetic_dataset):
...
+        UnischemaField('test_field', np.int32, (), ScalarCodec(IntegerType()), False),
+    ])
+
+    def test_row_generator(x):
+        """Returns a single entry in the generated dataset."""
+        return {'id': x,
+                'test_field': x*x}
+
+    rowgroup_size_mb = 256
+    dataset_url = "file:///tmp/partitioned_test_dataset"
```
selitvin: This will make the test leave garbage directory once its done. We should use `tmpdir` fixture that is automatically provided for us by the framework. E.g. of usage: `test_simple_read_moved_dataset` test in test_end_to_end.

petastorm/tests/test_predicates.py
```diff
@@ -124,3 +131,51 @@ def test_predicate_on_single_column(synthetic_dataset):
...
+
+        rows_rdd = sc.parallelize(range(rows_count))\
+            .map(test_row_generator)\
+            .map(lambda x: dict_to_spark_row(TestSchema, x))
+
+        spark.createDataFrame(rows_rdd, TestSchema.as_spark_schema()) \
+            .coalesce(10) \
+            .write \
+            .partitionBy('id') \
+            .mode('overwrite') \
```
selitvin: Let's remove unnecessary write options: don't need: coalesce, and mode.

petastorm/tests/test_predicates.py
```diff
@@ -124,3 +131,51 @@ def test_predicate_on_single_column(synthetic_dataset):
...
+            .parquet(dataset_url)
+
+    with make_reader(dataset_url, predicate=in_lambda(['id'], lambda x: x == 3)) as reader:
+        for sample in reader:
```
selitvin: One sample is enough to read. It will shorten the test run time.
(`next(reader)`)

petastorm/tests/test_predicates.py
```diff
@@ -124,3 +131,51 @@ def test_predicate_on_single_column(synthetic_dataset):
...
+            assert sample.id == 3
+    with make_reader(dataset_url, predicate=in_lambda(['id'], lambda x: x == '3')) as reader:
+        try:
+            # Predicate should have selected none, so a StopIteration should be raised.
+            next(reader)
```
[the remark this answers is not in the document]
jsgoller1: 👍

petastorm/tests/test_predicates.py
jsgoller1: Fixed.

petastorm/tests/test_predicates.py
jsgoller1: Fixed.

PR #310, opened 2019-02-11
#287 - fix materialize_dataset() serialization error by serializing fs factory

jsgoller1:
## Issue
#287 

## Problem
When using the `materialize_dataset()` context manager, the cleanup steps attempts to write rowgroup metadata to the generated files via `_generate_num_row_groups_per_file()`. This method attempts to serialize `pyarrow_filesystem` object, which works if the spark driver and executors reference the same JVM process (i.e. if they are on the same machine). However, if multiple processes are used, the `pyarrow_filesystem` object will contain invalid references. 

## Solution
Instead of `_generate_num_row_groups_per_file()` serializing the filesystem object, it should serialize a method for recreating the filesystem object on the spark executor, which can then be called to correctly reference files and obtain metadata.

### DIFF
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -50,7 +50,7 @@ class PetastormMetadataGenerationError(Exception):
 
 @contextmanager
 def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_summary_metadata=False,
-                        pyarrow_filesystem=None):
+                        filesystem_factory=None):
     """
     A Context Manager which handles all the initialization and finalization necessary
     to generate metadata for a petastorm dataset. This should be used around your
@@ -96,13 +96,13 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_
     yield
 
     # After job completes, add the unischema metadata and check for the metadata summary file
-    if pyarrow_filesystem is None:
+    if filesystem_factory is None:
         resolver = FilesystemResolver(dataset_url, spark.sparkContext._jsc.hadoopConfiguration())
-        filesystem = resolver.filesystem()
+        filesystem_factory = resolver.filesystem_factory()
         dataset_path = resolver.get_dataset_path()
     else:
-        filesystem = pyarrow_filesystem
         dataset_path = urlparse(dataset_url).path
+    filesystem = filesystem_factory()
 
     dataset = pq.ParquetDataset(
         dataset_path,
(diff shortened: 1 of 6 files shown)


petastorm/etl/dataset_metadata.py
```diff
@@ -98,10 +98,14 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_
...
     if pyarrow_filesystem is None:
         resolver = FilesystemResolver(dataset_url, spark.sparkContext._jsc.hadoopConfiguration())
+        fs_factory = resolver.filesystem_factory()
         filesystem = resolver.filesystem()
         dataset_path = resolver.get_dataset_path()
     else:
+        # If we are given a filesystem, then the factory
+        # is that filesystem's constructor
         filesystem = pyarrow_filesystem
+        fs_factory = type(filesystem)
```
selitvin: This is not going to work in general case. The construction of the filesystem object might be taking parameters provided by user. I would expect user to pass a factory and not an instance for this to work properly.

petastorm/fs_utils.py
```diff
@@ -139,3 +139,9 @@ def filesystem(self):
         :return: The pyarrow filesystem object
         """
         return self._filesystem
+
+    def filesystem_factory(self):
+        def factory():
+            resolver = FilesystemResolver(self._dataset_url)
```
selitvin: The object might have been created with additional parameters, not only the url (`self, dataset_url, hadoop_configuration=None, connector=HdfsConnector, hdfs_driver='libhdfs3')`). We need to reliably reconstruct this object with proper parameters.

Would be nice to see a unit test confirming this information is not lost.

petastorm/tests/test_dataset_metadata.py
```diff
@@ -29,3 +34,30 @@ def test_get_schema_from_dataset_url_bogus_url():
...
+def test_serialize_filesystem_factory(tmpdir):
+    SimpleSchema = Unischema('HelloWorldSchema', [
+        UnischemaField('id', np.int32, (), ScalarCodec(IntegerType()), False),
+        UnischemaField('foo', np.int32, (), ScalarCodec(IntegerType()), False),
+    ])
+
+    class BogusFS(pyarrow.LocalFileSystem):
+        def __getstate__(self):
+            raise RuntimeError("can not serialize")
+        # pass
```
selitvin: Please remove

jsgoller1: Removed.

petastorm/etl/dataset_metadata.py
jsgoller1: Changed.

petastorm/fs_utils.py
jsgoller1: Removed this - the factory methods now don't re-execute `FilesystemResolver`'s constructor.

PR #313, opened 2019-02-14
Don't read unneeded columns when working with ngrams.

selitvin:
Fixing a bug that caused all columns of a dataset to be read when `schema_fields=NGram(...)` was used.
Now, we collect all UnischemaField instances that appears in at least one of the time offsets within an NGram and load only corresponding columns.


mohanaprasad1994:
looks good

PR #315, opened 2019-02-15
Allow for unicode strings to be passed as regex filters when selectin…

selitvin:
…g which columns to read.

was previously failing with Python 2.7 when were passing `u'some_field.*'`.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -25,6 +25,7 @@ from pyspark import Row
 from pyspark.sql.types import StringType, ShortType, LongType, IntegerType, BooleanType, DoubleType, ByteType, \
     FloatType, DecimalType, DateType, TimestampType
 from pyspark.sql.types import StructField, StructType
+from six import string_types
 
 from petastorm.codecs import ScalarCodec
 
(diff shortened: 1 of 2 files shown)


petastorm/unischema.py
```diff
@@ -132,8 +133,12 @@ def create_schema_view(self, fields):
         """
 
         # Split fields parameter to regex pattern strings and UnischemaField objects
-        regex_patterns = [f for f in fields if isinstance(f, str)]
-        unischema_field_objects = [f for f in fields if not isinstance(f, str)]
+        regex_patterns = [f for f in fields if isinstance(f, string_types)]
+        unischema_field_objects = [f for f in fields if isinstance(f, UnischemaField)]
```
jsgoller1: Per the title, I assume that beforehand we were only allowing for `str`s, but now `string_types` should include Unicode strings?

jsgoller1:
Approved assuming Travis build succeeds.

petastorm/unischema.py
selitvin: Yep.

PR #316, opened 2019-02-15
Added a warning message: `make_batch_reader` supports only native Par…

selitvin:
…quet scalar/array types.

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -14,6 +14,7 @@
 
 import collections
 import logging
+import warnings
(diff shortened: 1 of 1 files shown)


selitvin:
> Just so I understand, it looks like if `get_schema_from_dataset_url()` succeeds, a warning should be issued - is the expected behavior here that `get_schema_from_dataset_url()` should raise an exception?
Yes. `make_batch_reader` currently supports only native parquet types ("non-Petastorm datasets") that do not require custom decoding. If `get_schema_from_dataset_url` succeeds, this means the dataset has unischema and is a Petastorm dataset.

PR #322, opened 2019-03-06
ARROW-1644/1599:  Excludes list inside structs fields from schema inference.

remysaissy:
Due to this pyarrow bug, we cannot use some fields in our datasets.
However, since these fields are not always mandatory for the use case to be implemented and since data duplication would be too costly, we can afford to ignore it directly when loading the parquet file.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -17,10 +17,13 @@ in several different python libraries. Currently supported are pyspark, tensorfl
 """
 import copy
 import re
+import warnings
 from collections import namedtuple, OrderedDict
 from decimal import Decimal
 
 import numpy as np
+from pyarrow.lib import ListType
+from pyarrow.lib import StructType as pyStructType
 from pyspark import Row
 from pyspark.sql.types import StringType, ShortType, LongType, IntegerType, BooleanType, DoubleType, ByteType, \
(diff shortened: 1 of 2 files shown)


selitvin:
Sorry, was away for awhile. Will follow up on your PRs today.

petastorm/unischema.py
```diff
@@ -234,9 +234,10 @@ def from_arrow_schema(cls, parquet_dataset):
         for column_name in arrow_schema.names:
             arrow_field = arrow_schema.field_by_name(column_name)
             field_type = arrow_field.type
-            codec, np_type = _numpy_and_codec_from_arrow_type(field_type)
+            if not isinstance(field_type, ListType):
```
selitvin: I think this discards valid cases that are supported. That's the reason for failure of the tests. We should be more precise here. Would also consider to issue a `warning.warn` for the user not to be surprised. Would need to cover this case with a unit test.

Please let us know if you plan to continue working on it (would be more than happy to get this as a contribution). Alternatively, we can help resolve this case.

Thanks!

remysaissy: Good idea, I'll add this warning.
What do you mean by more precise?

Thanks for the proposal, I still can allocate time to work on it so I will continue for now. I'll let you know if needed.

PR #323, opened 2019-03-06
Unischema supports Parquet schema with more than 255 fields

remysaissy:
Many of our datasets have more than 255 fields.
This commit provides an alternative namedtuple implementation 'namedtuple2' to support more than 255 fields with the Python 3.6 interpreter.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,0 +1,265 @@
+Petastorm includes derived work from CPython(https://github.com/python/cpython/blob/3.6/Lib/collections/__init__.py). The derived work includes:
+
+* changes to support more than 255 fields
+** change the __new__ method implementation
+** add a __reduce__ method to make it pickable
+
+* changes to match the pylint 2.7 checker
+** add of warning ``#pylint: disable=W0122`` for the class_definition exec() call
+** adapt the verbose print() call
+
+CPython is under the Python License 2.0:
+A. HISTORY OF THE SOFTWARE
+==========================
+
+Python was created in the early 1990s by Guido van Rossum at Stichting
+Mathematisch Centrum (CWI, see http://www.cwi.nl) in the Netherlands
+as a successor of a language called ABC. Guido remains Python's
+principal author, although it includes many contributions from others.
+
+In 1995, Guido continued his work on Python at the Corporation for
+National Research Initiatives (CNRI, see http://www.cnri.reston.va.us)
+in Reston, Virginia where he released several versions of the
+software.
+
+In May 2000, Guido and the Python core development team moved to
+BeOpen.com to form the BeOpen PythonLabs team. In October of the same
+year, the PythonLabs team moved to Digital Creations, which became
+Zope Corporation. In 2001, the Python Software Foundation (PSF, see
+https://www.python.org/psf/) was formed, a non-profit organization
+created specifically to own Python-related Intellectual Property.
+Zope Corporation was a sponsoring member of the PSF.
+
+All Python releases are Open Source (see http://www.opensource.org for
+the Open Source Definition). Historically, most, but not all, Python
+releases have also been GPL-compatible; the table below summarizes
+the various releases.
+
+ Release Derived Year Owner GPL-
+ from compatible? (1)
+
+ 0.9.0 thru 1.2 1991-1995 CWI yes
+ 1.3 thru 1.5.2 1.2 1995-1999 CNRI yes
+ 1.6 1.5.2 2000 CNRI no
+ 2.0 1.6 2000 BeOpen.com no
+ 1.6.1 1.6 2001 CNRI yes (2)
+ 2.1 2.0+1.6.1 2001 PSF no
+ 2.0.1 2.0+1.6.1 2001 PSF yes
+ 2.1.1 2.1+2.0.1 2001 PSF yes
+ 2.1.2 2.1.1 2002 PSF yes
+ 2.1.3 2.1.2 2002 PSF yes
+ 2.2 and above 2.1.1 2001-now PSF yes
+
+Footnotes:
+
+(1) GPL-compatible doesn't mean that we're distributing Python under
+ the GPL. All Python licenses, unlike the GPL, let you distribute
+ a modified version without making your changes open source. The
+ GPL-compatible licenses make it possible to combine Python with
+ other software that is released under the GPL; the others don't.
+
+(2) According to Richard Stallman, 1.6.1 is not GPL-compatible,
+ because its license has a choice of law clause. According to
+ CNRI, however, Stallman's lawyer has told CNRI's lawyer that 1.6.1
+ is "not incompatible" with the GPL.
+
+Thanks to the many outside volunteers who have worked under Guido's
+direction to make these releases possible.
+
+
+B. TERMS AND CONDITIONS FOR ACCESSING OR OTHERWISE USING PYTHON
+===============================================================
+
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+--------------------------------------------
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
+("PSF"), and the Individual or Organization ("Licensee") accessing and
+otherwise using this software ("Python") in source or binary form and
+its associated documentation.
+
+2. Subject to the terms and conditions of this License Agreement, PSF hereby
+grants Licensee a nonexclusive, royalty-free, world-wide license to reproduce,
+analyze, test, perform and/or display publicly, prepare derivative works,
+distribute, and otherwise use Python alone or in any derivative version,
+provided, however, that PSF's License Agreement and PSF's notice of copyright,
+i.e., "Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019 Python Software Foundation;
+All Rights Reserved" are retained in Python alone or in any derivative version
+prepared by Licensee.
+
+3. In the event Licensee prepares a derivative work that is based on
+or incorporates Python or any part thereof, and wants to make
+the derivative work available to others as provided herein, then
+Licensee hereby agrees to include in any such work a brief summary of
+the changes made to Python.
+
+4. PSF is making Python available to Licensee on an "AS IS"
+basis. PSF MAKES NO REPRESENTATIONS OR WARRANTIES, EXPRESS OR
+IMPLIED. BY WAY OF EXAMPLE, BUT NOT LIMITATION, PSF MAKES NO AND
+DISCLAIMS ANY REPRESENTATION OR WARRANTY OF MERCHANTABILITY OR FITNESS
+FOR ANY PARTICULAR PURPOSE OR THAT THE USE OF PYTHON WILL NOT
+INFRINGE ANY THIRD PARTY RIGHTS.
+
+5. PSF SHALL NOT BE LIABLE TO LICENSEE OR ANY OTHER USERS OF PYTHON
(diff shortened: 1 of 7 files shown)


petastorm/tf_utils.py
```diff
@@ -384,3 +384,171 @@ def dequeue_sample_impl():
         return named_tuple_dataset
     else:
         raise NotImplementedError('make_petastorm_dataset does not support NGram yet.')
+
```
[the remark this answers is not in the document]
remysaissy: ok thanks.

The boilerplate of the namedtuple2 code comes from a mix between a 'lifted' variation of python 3.6 namedtuple gist that I found clearer:
https://gist.github.com/malcolmgreaves/a8ff32e814053f805ba9c0aae380771c
This is the Python license I guess.

The >255 support lies in the __new__ method that I've implemented myself for this PR.
We have put this code in the criteo-forks/petastorm but temporarly, until the feature is merged into your repository so no particular license on that part.

petastorm/tf_utils.py
```diff
@@ -384,3 +384,171 @@ def dequeue_sample_impl():
...
+"""
+
+_repr_template = '{name}=%r'
+
+_field_template = '''\
+    {name} = _property(_itemgetter({index}), doc='Alias for field number {index}')
+'''
+
+
+def namedtuple2(typename, field_names, *, verbose=False, rename=False, module=None):
```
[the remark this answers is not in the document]
remysaissy: I was out of ideas on that. namedtuple_gt_255_fields sounds clear enough, let's use it!

selitvin:
Tried #330 (read older datasets) with your PR and seems to work.

When using this PR, I get an error when using Python 2.7:
```
$ pytest petastorm/tests/test_end_to_end.py 
ImportError while loading conftest '/home/yevgeni/uatc/dataset-toolkit/petastorm/tests/conftest.py'.
petastorm/__init__.py:15: in <module>
    from petastorm.reader import make_reader, make_batch_reader  # noqa: F401
petastorm/reader.py:25: in <module>
    from petastorm.etl import dataset_metadata, rowgroup_indexing
petastorm/etl/dataset_metadata.py:29: in <module>
    from petastorm.unischema import Unischema
petastorm/unischema.py:31: in <module>
    from petastorm.tf_utils import namedtuple_gt_255_fields
E   SyntaxError: unqualified exec is not allowed in function 'namedtuple_gt_255_fields' it contains a nested function with free variables (tf_utils.py, line 528)
```
I don't understand why Travis does not fail in this case? I think to fix this we would need to move python3 named tuple into its own module, distinct from python 2 module.

As for the origin of the code, we might need to add PSF license header or something. I will figure out the licensing issue.

petastorm/tf_utils.py
```diff
@@ -384,3 +384,171 @@ def dequeue_sample_impl():
         return named_tuple_dataset
     else:
         raise NotImplementedError('make_petastorm_dataset does not support NGram yet.')
+
+
```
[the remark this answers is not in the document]
remysaissy: done using the six module.

remysaissy:
@selitvin , the PSF should not be needed now.
I've realized that in Python it is possible to simply override the __new__ method of the namedtuple to fix the 255 fields limit.

Looks like the checks have all passed.
Let me know if you think we should improve them further.

selitvin:
If hacking namedtuple all the way through would prove too cumbersome, we may consider not using namedtuple here all together. It seems to be too magical class. Maybe we should come up with a custom container class  (just match `_asdict` and `_fields` properties since it is being used by existing code)?

remysaissy:
@selitvin thanks, my bad sorry I didn't realize that those tests were disabled. Thanks for pointing out.

 - The previous implementation was fine but had performance issues
 - The current, simpler implementation actually doesn't work because the __new__ method should actually be rewritten before namedtuple is called, which we can do only using the previous implementation

I agree with the custom container solution. It is probably best this way.

selitvin:
Yeah. Implementing a custom container could be a bit more evolved thing to do. Don't know if you are up to it. I think it's also ok to use your original proposal of modified namedtuple. I assume you used it in your production code and it worked well for you. If you choose to go that route, I prepared a NOTICE file (https://github.com/selitvin/petastorm/tree/python_license_notice) that we would need to include to cover the legal aspects of pasting modified code from Python 3 libraries. We should just add summary of modifications instead of 
```
TODO:
<<symmary of the modifications>>
```

remysaissy:
@selitvin thanks, sorry I didn't realize that those tests were disabled. Thanks for pointing out.

ok let's do this. I've updated the PR to reuse the previous implementation in a separate module and added the NOTICE file.

remysaissy:
@selitvin , https://travis-ci.com/uber/petastorm/jobs/185738409 is the error I keep having with the namedtuple based implementation.
I am not sure of the reason behind this failure.
Do you have an idea?

remysaissy:
@selitvin the error in the CI is the issue I keep having with the namedtuple, python based implementation.
It occurs in a benchmark test but I am not sure of its root cause yet.
If you have feedback on it, I'll welcome it :).

selitvin:
I see failures in the following cases:
- Either pyarrow serialization is used to serialize our custom namedtuples
- PySpark tries to serialize namestuple (https://jira.apache.org/jira/browse/SPARK-22674).

namedtuple is just a syntactic sugar that is not really needed and turns out to be pretty poisonous. 
How about the following idea: we make a switch that would result in using dictionaries as the return type.

Probably should shoot for deprecation of namedtuples as a reader return at some stage. It will be off by default for now.

What do you think?

remysaissy:
Agree.
and the property-like sugar of namedtuple can be relatively easily re-implemented too if we want.

I can do it as part of this PR.

remysaissy:
@selitvin  some news as I've investigated into this RLock issue a bit more.
It appears that it is not related to namedtuples with arrow but rather to the WorkerType used.
The PROCESS workertype triggers the failure, whereas using THREAD does not.

I've came across https://stackoverflow.com/questions/44144584/typeerror-cant-pickle-thread-lock-objects which gives some interesting insights on what's happening (or I guess so).
So it might not be necessary to remove the namedtuple, depending on the outcome of it :).

selitvin:
From the failing tests, it seems like there are many software pieces that don't want us to use a custom tuple :) 

What do you say we give a shot to optionally dropping the syntax sugar and returning a dictionary. And I would setup a goal to gradually deprecate namedtuple return type going forward? I think it solves your problem. I can create a PR for that.

remysaissy:
I think the issue is fixed for the RLock. The custom namedtuple was missing the __reduce__ method required to properly pickle.

The current error looks related to another commit, can you confirm?

selitvin:
> I think the issue is fixed for the RLock. The custom namedtuple was missing the **reduce** method required to properly pickle.
> 
> The current error looks related to another commit, can you confirm?

Yes. My mistake. Will fix.

selitvin:
Awesome - glad that you nailed it! Let me do another final pass on this today or tomorrow.

NOTICE
```diff
@@ -0,0 +1,265 @@
+Petastorm includes derived work from CPython(https://github.com/python/cpython/blob/master/Lib/collections/__init__.py#L301). The derived work includes:
```
selitvin: This link points to a newer version of cpython. Its implementation of namedtuple is significantly different from the ones we have here.
This one is closer: https://github.com/python/cpython/blob/2.7/Lib/collections.py (but still not exactly the same). I wonder why are the differences?

petastorm/namedtuple.py
```diff
@@ -0,0 +1,202 @@
...
+        return self.__class__.__name__ + '({repr_fmt})' % self
+
+    def _asdict(self):
+        'Return a new OrderedDict which maps field names to their values.'
+        return OrderedDict(zip(self._fields, self))
+
+    def __getnewargs__(self):
+        'Return self as a plain tuple. Used by copy and pickle.'
+        return tuple(self)
+
```
selitvin: We don't have `__dict__`. Is this on purpose?
https://github.com/python/cpython/blob/6cbb4c0795099b79f0a7c7d20df4ba1c1ec0ac24/Lib/collections.py#L290

Also some parts of the "def namedtuple_gt_255_fields" are different from python 2.7 (I mean apart from the  the differences related to the problem we are trying to solve). Don't see anything critical, but still interested in the reason of the differences.

petastorm/namedtuple.py
```diff
@@ -0,0 +1,202 @@
...
+        """ Restore an object of namedtuple"""
+        k = (name, fields)
+        cls = __cls.get(k)
+        if cls is None:
+            cls = namedtuple_gt_255_fields(name, fields)
+            __cls[k] = cls
+        return cls(*value)
+
+    def _hack_namedtuple_gt_255_fields(cls):
+        """ Make class generated by namedtuple picklable """
```
selitvin: Can you please document this pickling hack?

petastorm/namedtuple.py
```diff
@@ -0,0 +1,202 @@
...
+import sys as _sys
+from collections import namedtuple
+from keyword import iskeyword as _iskeyword
+import six
+
+
+if six.PY2:
+    namedtuple_gt_255_fields = namedtuple
+else:
+    _class_template = """\
```
selitvin: Apparently Python 3.7 also supports >255 namedtuple fields. Let's use a standard namedtuple for 3.7.

NOTICE
remysaissy: Oh right, I was referencing the master branch here. I will reference this one then (or another one even closer if I find it).

I'm not sure about the reason for the differences, I think it might have to do with the performances of the interpreter between string interpolation to define a class vs writing the method ahead.

petastorm/namedtuple.py
remysaissy: Yes, it makes __dict__ being the fields of the namedtuple with its values only.


The other small differences are because I based this work on another commit of collections.py (but ref it as master branch so...).

To make things clearer for the NOTICE, I can rebase it against the 2.7 implementation, that will make the NOTICE and the derived work changes more compact and focused only on the > 255 fields support.

petastorm/namedtuple.py
remysaissy: Ok

petastorm/namedtuple.py
remysaissy: Not on purpose but it was not necessary in the end.

I will make a pass on the code in order to ensure it sticks as much as possible to the cython 2.7 implementation since the diffs are more than what is necessary.

NOTICE
remysaissy: My implementation derives from this one: https://github.com/python/cpython/blob/3.6/Lib/collections/__init__.py
I will update the NOTICE to mention the 3.6 branch instead of the master one.

petastorm/namedtuple.py
remysaissy: Indeed, if I remember well, the fields limit is maxint in python 3.7. I'll limit the use to Python 3.0.x to 3.6.x.

petastorm/namedtuple.py
remysaissy: It comes from 3.6 implementation initially where the __dict__ is not redefined.

selitvin:
Thank you very much for all the hard work! Let's ship it.

Last question before we land this: you removed the serialization hack. We don't need it anymore?

remysaissy:
@selitvin, you're welcome.
The pickling support is still here, I just integrated it properly in the generated class.
 - the `__reduce__` method is part of the definition of the class template instead of being injected after the class creation
 - the `_restore_namedtuple_gt_255_fields` cannot be generated so I've left it outside (after a few tests with the `__setstate__` method that should be doing that restoration but I couldn't get it to work properly in our specific case)
 - I removed the _cls = {} variable since I realized that while caching previously generated namedtuple  may be nice, this dictionary is actually never cleaned up so we could end up with extra memory usage

PR #324, opened 2019-03-07
Adding unit tests to verify that Petastorm can work with existing parquet stores with > 255 columns.

selitvin:
Python 3 does not support namedtuples with > 255 fields. #323 attempts to solve this issue.
These tests can be used to verify that the fix works. Currently the tests are disabled for Python 3
and we need to reenable them once the solution is in place.

### DIFF
--- a/petastorm/tests/test_tf_dataset.py
+++ b/petastorm/tests/test_tf_dataset.py
@@ -18,16 +18,17 @@ from copy import copy
 
 import numpy as np
 import pytest
+import six
 import tensorflow as tf
 
-from petastorm import make_reader
+from petastorm import make_reader, make_batch_reader
 from petastorm.ngram import NGram
 from petastorm.predicates import in_lambda
 from petastorm.tests.test_common import TestSchema
 from petastorm.tf_utils import make_petastorm_dataset
 
 _EXCLUDE_FIELDS = set(TestSchema.fields.values()) \
-    - {TestSchema.matrix_nullable, TestSchema.string_array_nullable, TestSchema.decimal}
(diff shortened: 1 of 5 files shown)


petastorm/tests/test_tf_dataset.py
```diff
@@ -150,3 +151,24 @@ def test_dataset_on_ngram_not_supported(synthetic_dataset, reader_factory):
...
+        # Make sure we have static shape info for all fields
+        for shape in dataset.output_shapes:
+            # TODO(yevgeni): check that the shapes are actually correct, not just not None
+            assert shape.dims is not None
+
+        # Read a bunch of entries from the dataset and compare the data to reference
+        with tf.Session() as sess:
+            iterator = iterator.get_next()
+            sample = sess.run(iterator)._asdict()
+            assert sorted(sample.keys()) == sorted(many_columns_non_petastorm_dataset.data[0].keys())
```
SashaOv: Could `set(l)==set(r)` instead of `sorted()` be faster ?

selitvin:
@jsgoller1 
> @selitvin Just so I understand: we're adding this test knowing that the issue won't arise in Python 2.7, but since we don't have a solution in place for 3.2 yet, we're skipping it with the intent to re-enable once we have a working solution?

Yes. Wanted to make sure we have a way to test #323

petastorm/tests/test_tf_dataset.py
selitvin: Done.

selitvin:
> LGTM. Not sure where do we use tuples in the code?

Unfortunately we do (will not make this mistake again :) ). The values returned from the reader are namedtuple. The goal was to enable auto-completion, when developing under a live interpreter.

PR #325, opened 2019-03-10
Make hdfs connect arguments unicode strings

selitvin:
pyarrow==0.12.1 assume hdfs connect arguments are unicode. Make sure we pass unicode.
(tested using python 2/3 and pyarrow 0.12.1 and 0.11.1)

### DIFF
--- a/petastorm/hdfs/namenode.py
+++ b/petastorm/hdfs/namenode.py
@@ -16,9 +16,11 @@ import functools
 import inspect
 import logging
 import os
+from distutils.version import LooseVersion
 from xml.etree import ElementTree as ET
 
 import pyarrow
+import six
 from pyarrow.hdfs import HadoopFileSystem
 from pyarrow.lib import ArrowIOError
 from six.moves.urllib.parse import urlparse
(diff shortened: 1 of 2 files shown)


petastorm/hdfs/namenode.py
```diff
@@ -247,7 +249,13 @@ def hdfs_connect_namenode(cls, url, driver='libhdfs3'):
         :param driver: An optional driver identifier
         :return: Pyarrow HDFS connection object.
         """
-        return pyarrow.hdfs.connect(url.hostname or 'default', url.port or 8020, driver=driver)
+        if LooseVersion(pyarrow.__version__) < LooseVersion('0.12.0'):
+            hostname = url.hostname or 'default'
+            driver = driver
+        else:
+            hostname = six.text_type(url.hostname or 'default')
```
SashaOv: Not sure I understand the 'default' thing

selitvin: Added a comment:
```python
# According to pyarrow.hdfs.connect:
#    host : NameNode. Set to "default" for fs.defaultFS from core-site.xml
# So we pass 'default' as a host name if the url does not specify one (i.e. hdfs:///...)
```

PR #326, opened 2019-03-10
Convert `datetime.date` to int64 (nanosec from epoch)

selitvin:
`pyarrow==0.12.1` now reads parquet `DateType` columns as python datetime object.
TF does not support conversion of this native python time. We convert datetime.date column
to the number of nano-seconds since epoch in int64 to make the column into TF friendly.

### DIFF
--- a/petastorm/tests/test_tf_utils.py
+++ b/petastorm/tests/test_tf_utils.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 from __future__ import division
 
+import datetime
+from calendar import timegm
 from collections import namedtuple
 from contextlib import contextmanager
 from decimal import Decimal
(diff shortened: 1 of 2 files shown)


petastorm/tf_utils.py
```diff
@@ -69,13 +71,20 @@ def _sanitize_field_tf_types(sample):
...
             next_sample_dict[k] = v.astype(np.int32)
         elif isinstance(v, np.ndarray) and v.dtype.type in (np.bytes_, np.unicode_):
             if v.size != 0:
                 next_sample_dict[k] = v.tolist()
+        elif isinstance(v, np.ndarray) and v.dtype.kind == 'O' and isinstance(v[0], datetime.date):
+            # Pyarrow 0.12.1 started returning python datetime.date when parquet column is a DateType() column.
+            # Convert values in such column into nsec from epoch int64.
+            def date_to_nsec_from_epoch(dt):
+                return timegm(dt.timetuple()) * 1000000000
+            to_nsec = np.vectorize(date_to_nsec_from_epoch)
```
SashaOv: This is executed for every value. Since this is constant, it would make sense to define it outside of this function

petastorm/tests/test_tf_utils.py
```diff
@@ -46,13 +48,17 @@ def _tf_session():
         coord.join(threads)
 
 
-def test_empty_dict():
-    # Check two types that should be promoted/converted (uint16 and Decimal) and one that should not be
-    # modified (int32)
+def test_sanitize_field_tf_types():
+    expected_datetime_array = [datetime.date(1970, 1, 1), datetime.date(2015, 9, 29)]
+    expected_datetime_ns_from_epoch = [timegm(dt.timetuple()) * 1000000000 for dt in expected_datetime_array]
```
SashaOv: Wanna add a test with non-zero nanoseconds?

selitvin: I don't think datetime gets to the nanoseconds resolution...

Issue #327, opened 2019-03-13
UnischemaField name field cannot set `name` value 


wangqiaoshi:
code:
test_schema = Unischema('test_schema', [
    ...:         UnischemaField('id', np.int32, (), ScalarCodec(IntegerType()), False),
    ...:         UnischemaField('name', np.string, (), ScalarCodec(StringType()), False)
    ...:     ])

error:
 114         # Generates attributes named by the field names as an access syntax sugar.
    115         for f in fields:
--> 116             setattr(self, f.name, f)
    117 
    118     def create_schema_view(self, fields):

AttributeError: can't set attribute

selitvin:
This is due to a conflict with the `name` property of the `Unischema` object itself. 
Dynamic property names were created as a syntax sugar and is widely used in our codebase.

Suggest the following solution:
- Rename current `name` property to `_name`. This pattern is used by `collections.namedtuple`, I suspect for the same reason. The rename would reduce a probability of a conflict since `name` is a common column name.
- Issue a warning if a conflict is detected with other properties (e.g. `fields`). Don't create a dynamic property and don't raise an error (just warning).

selitvin:
Created PR #328 to address this issue

selitvin:
77727b0b5be2b4587eb2a10f570e8849ec6122f8 properly solves this issue.

PR #328, opened 2019-03-13
Rename `Unischema.name` to `Unischema._name` property to avoid conflict user fields named `name`.

selitvin:
"name" is a likely name for a user column. Existing `Unishema.name` attribute conflicts with
dynamically generated attributes. We rename `Unischema.name` into `Unischema._name`. If there
is a conflict between user field and existing attribute, we emit a warning and skip dynamic
attribute creation for that field.


selitvin:
Solves #327

PR #330, opened 2019-03-14
Adding legacy dataset read test

selitvin:
The commit includes datasets generated using `generate_dataset_for_legacy_test` script for several legacy versions.


selitvin:
> @selitvin - why are we committing the dataset files to the repo? Is it not possible for a user to regenerate them?

@jsgoller1 
These datasets were generated using older versions of petastorm. In order to generate them you would need to checkout a different code base. While it is possible, not sure how complicated would it be to do on Travis. Do you have an easy way to achieve this?

PR #331, opened 2019-03-14
Add an option to warn and ignore unsupported columns

remysaissy:
Some Criteo parquet datasets contain unsupported fields.
The current behaviour of Petastorm is to raise a ValueError in the from_arrow_schema()::_numpy_and_codec_from_arrow_type() method when such a field is detected.
This patch introduces an optional argument to print only a warning and ignore the field instead of raising an exception, thus enabling the use of the 'partial' dataset.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -227,7 +227,7 @@ class Unischema(object):
         When the warn_only parameter is turned to True, unsupported column types prints only warnings.
 
         :param arrow_schema: :class:`pyarrow.lib.Schema`
-        :param omit_unsupported_fields: ::class: `Boolean`
+        :param omit_unsupported_fields: :class:`Boolean`
         :return: A :class:`Unischema` object.
         """
         meta = parquet_dataset.pieces[0].get_metadata(parquet_dataset.fs.open)
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -25,7 +25,8 @@ import numpy as np
 from pyarrow.lib import ListType
 from pyarrow.lib import StructType as pyStructType
 from pyspark import Row
-from pyspark.sql.types import StringType, ShortType, LongType, IntegerType, BooleanType, DoubleType, ByteType, \
+from pyspark.sql.types import StringType, ShortType, LongType, IntegerType, BooleanType, DoubleType, \
+    ByteType, \
     FloatType, DecimalType, DateType, TimestampType
 from pyspark.sql.types import StructField, StructType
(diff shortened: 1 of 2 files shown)


petastorm/unischema.py
```diff
@@ -215,13 +217,15 @@ def make_namedtuple_tf(self, *args, **kargs):
         return self._get_namedtuple()(*args, **kargs)
 
     @classmethod
-    def from_arrow_schema(cls, parquet_dataset):
+    def from_arrow_schema(cls, parquet_dataset, warn_only=False):
```
selitvin: How about  `warn_only` -> `omit_unsupported_fields` (or some variant of the name to indicate which warnings are you referring to ) ?

petastorm/unischema.py
```diff
@@ -234,9 +238,15 @@ def from_arrow_schema(cls, parquet_dataset):
         for column_name in arrow_schema.names:
             arrow_field = arrow_schema.field_by_name(column_name)
             field_type = arrow_field.type
-            codec, np_type = _numpy_and_codec_from_arrow_type(field_type)
-
-            unischema_fields.append(UnischemaField(column_name, np_type, (), codec, arrow_field.nullable))
+            try:
+                codec, np_type = _numpy_and_codec_from_arrow_type(field_type)
```
selitvin: Please reduce the scope of try/except to the line that raises the error to avoid wrong behavior if line 243 raises `ValueErorr`

petastorm/unischema.py
```diff
@@ -234,9 +238,15 @@ def from_arrow_schema(cls, parquet_dataset):
...
+                unischema_fields.append(UnischemaField(column_name, np_type, (), codec, arrow_field.nullable))
+            except ValueError as ex:
+                if warn_only:
+                    warnings.warn('Column %r has an unsupported field %r. Ignoring...'
+                                  % (column_name, field_type))
+                else:
+                    raise ex
```
selitvin: `raise ex` would drop inner call stack information. 
```python
except VaiueError:
   ...
   raise
```
would preserve it.

petastorm/unischema.py
remysaissy: good idea, that will be much clearer.

petastorm/unischema.py
remysaissy: Ok

petastorm/unischema.py
remysaissy: indeed, I change it.

PR #336, opened 2019-03-26
Dummy spark codec for skipping encoding/decoding on spark datasets

mike-grayhat:
Right now `UnischemaField` expects a codec, however if i have a spark dataset having a column of dtype `array<array<string>>` there is no easy way to create `UnischemaField` for it or at least it's not well documented how to do so. With this codec it's quite easy to define shema field and get nested arrays from petastorm reader later:

```python
from pyspark.sql import types as T

# array<array<string>> spark field
my_field = UnischemaField(
    name, np.unicode_, (None, None),
    NoopCodec(T.ArrayType(T.ArrayType(T.StringType()))),
    False
)
```

### DIFF
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -170,7 +170,7 @@ class CompressedNdarrayCodec(DataframeColumnCodec):
         return BinaryType()
 
 
-class DummySparkCodec(DataframeColumnCodec):
+class NoopCodec(DataframeColumnCodec):
     """Uses internal arrow data representation without encoding/decoding"""
 
     def __init__(self, spark_type):
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -170,6 +170,26 @@ class CompressedNdarrayCodec(DataframeColumnCodec):
         return BinaryType()
 
 
+class DummySparkCodec(DataframeColumnCodec):
+    """Uses internal arrow data representation without encoding/decoding"""
+    
+    def __init__(self, spark_type):
+        """Constructs a codec.
+
+        :param spark_type: an instance of a Type object from :mod:`pyspark.sql.types`
+        """
+        self._spark_type = spark_type
+    
+    def encode(self, unischema_field, value):
+        return value
+
+    def decode(self, unischema_field, value):
+        return value
+
+    def spark_dtype(self):
+        return self._spark_type
+
+
 class ScalarCodec(DataframeColumnCodec):
     """Encodes a scalar into a spark dataframe field."""
 
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -172,14 +172,14 @@ class CompressedNdarrayCodec(DataframeColumnCodec):
 
 class DummySparkCodec(DataframeColumnCodec):
     """Uses internal arrow data representation without encoding/decoding"""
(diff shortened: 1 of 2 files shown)


selitvin:
Thank you for the PR!

Would like to get a little more context:

- Is the dataset you are reading created using Petastorm (you used `with materialize_dataset(...)` or this is some existing Parquet store? 
- Are you reading the dataset using `make_batch_reader`? 

We have added a check in `0.7.0` that ignores columns that are "list of lists" or "list of structures" types (#322). I assume that's what you are trying to accomplish by this PR?

mike-grayhat:
I got the impression that `make_reader` is favorable approach of working with `petastorm`. My workflow is quite simple - i vectorize my data with spark and then save it with petastorm metadata:
```python
df = spark.table('my_table')
df_vec = some_etl(df)
with materialize_dataset(...):
    df_vec.write.parquet(output_url)
```
And yes, my case is around nlp so i need nested structures. I know that pyarrow doesn't support them yet and i'm quite happy that i can get them with `petastorm`.

selitvin:
QQ: what do you do with the variable list data when you get it? Are you using it from Tensorflow, PyTorch or pure python? Are you using trasform_spec to convert it into some tensor form?

I think this is a very interesting use case. Indeed, you should use `make_reader`. Note that pyarrow does support list of lists in `0.12.0` (we rely on pyarrow to do IO). 

Maybe we should consider a different name? IMO "Dummy" does not convey well enough what we are trying to achieve with it. Maybe 'NoopCodec' or 'PassthroughCodec'?

We should add some test coverage for that. Some trivial test to test_codecs.py, but we would need an integration test. You can add list of lists to TestSchema in `test_common.py`.  The following would need to be taken into account:
- The new field needs to be excluded in TF tests since the type won't be compatible with native TF types. All TF tests has mechanism that excludes a field since some fields are nullable and as such are not compatible with TF tensors.
- Tests that compare values would probably need special handling to compare list of lists. I think we use mostly `np.testing.assert_equal`, which I think won't be able to handle this type.

Please let me know if you can put some time into extending the tests. If not, I can probably help you with that.

mike-grayhat:
I'm using this data to train pytorch and crfsuite models (nested list represents token features of a sentence). Transform spec causes errors for me in some cases, so i'm trying to avoid any transformations during data loading. For me it's not essential to have nested lists, since i'm not using them in pytorch model, only in crfsuite, but it's very convenient to have one source of data for all models. Anyway as far as i understand even for simple lists i have to use such codec. 
I don't mind changing codec name so `NoopCodec` is fine with me. I'll look into tests, but it's possible i will need some help with them.

selitvin:
Cool. This all looks good to me. Can you please update the name to `NoopCodec` and I will land it right after I am done with 0.7.1 release? I can help you with the tests once this code is on master.

selitvin:
BTW, curious, does excluding that particular column using `schema_fields=` argument of `make_reader` would also be a viable solution?

mike-grayhat:
Hi. Sorry, i was quite busy these days, most likely will push the updates this weekend. Yes, i usually just omit columns i don't want in schema fields.

selitvin:
Please reopen if still needed.

PR #338, opened 2019-03-27
Make user aware of a field name that failed decoding of a row

selitvin:
Raise DecodeFieldError exception. Added tests for `decode_row` function.

### DIFF
--- a/petastorm/tests/test_decode_row.py
+++ b/petastorm/tests/test_decode_row.py
@@ -1,0 +1,41 @@
+#  Copyright (c) 2017-2018 Uber Technologies, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


petastorm/tests/test_decode_row.py
```diff
@@ -0,0 +1,41 @@
...
+
+from petastorm.codecs import NdarrayCodec
+from petastorm.unischema import UnischemaField, Unischema
+from petastorm.utils import decode_row, DecodeFieldError
+
+MatrixField = UnischemaField('matrix', np.float64, (10, 10), NdarrayCodec(), False)
+MatrixSchema = Unischema('TestSchema', [MatrixField])
+
+
+def test_nominal_case():
```
SashaOv: They also sometimes call this "happy path"

PR #339, opened 2019-03-27
Fix #333: If integer_object_nulls is set to False, nullable integer fields are return as floats

### LINKED ISSUE #333: Failure in case of a null value in an integer scalar field
If a dataset contains a nullable integer field with a null value, reading the dataset fails (using `make_reader`):

```
File "/lib/python_env/petastorm/workers_pool/dummy_pool.py", line 72, in get_results
    self._worker.process(*args, **kargs)
  File "/lib/python_env/petastorm/py_dict_reader_worker.py", line 142, in process
    lambda: self._load_rows(parquet_file, piece, shuffle_row_drop_partition))
  File "/lib/python_env/petastorm/cache.py", line 39, in get
    return fill_cache_func()
  File "/lib/python_env/petastorm/py_dict_reader_worker.py", line 142, in <lambda>
    lambda: self._load_rows(parquet_file, piece, shuffle_row_drop_partition))
  File "/lib/python_env/petastorm/py_dict_reader_worker.py", line 161, in _load_rows
    return [transform_func(utils.decode_row(row, self._schema)) for row in all_rows]
  File "/lib/python_env/petastorm/utils.py", line 60, in decode_row
    decoded_row[field_name] = codec.decode(schema.fields[field_name], row[field_name])
  File "/lib/python_env/petastorm/codecs.py", line 205, in decode
    return unischema_field.numpy_dtype(value)
```

selitvin:
The fix is to use appropriate switch in a call to pyarrow table to pandas conversion.

### DIFF
--- a/petastorm/py_dict_reader_worker.py
+++ b/petastorm/py_dict_reader_worker.py
@@ -226,11 +226,13 @@ class PyDictReaderWorker(WorkerBase):
             return filtered_decoded_predicate_rows
 
     def _read_with_shuffle_row_drop(self, piece, pq_file, column_names, shuffle_row_drop_partition):
+        # If integer_object_nulls is set to False, nullable integer fields are return as floats
+        # with nulls translated to nans
         data_frame = piece.read(
             open_file_func=lambda _: pq_file,
             columns=column_names,
             partitions=self._dataset.partitions
-        ).to_pandas()
+        ).to_pandas(integer_object_nulls=True)
 
         num_rows = len(data_frame)
         num_partitions = shuffle_row_drop_partition[1]
--- a/petastorm/tests/test_common.py
+++ b/petastorm/tests/test_common.py
@@ -50,6 +50,7 @@ TestSchema = Unischema('TestSchema', [
     UnischemaField('matrix_nullable', np.uint16, _DEFAULT_IMAGE_SIZE, NdarrayCodec(), True),
     UnischemaField('sensor_name', np.unicode_, (1,), NdarrayCodec(), False),
     UnischemaField('string_array_nullable', np.unicode_, (None,), NdarrayCodec(), True),
+    UnischemaField('integer_nullable', np.int32, (), ScalarCodec(ShortType()), True),
 ])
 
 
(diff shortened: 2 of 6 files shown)


petastorm/tests/test_common.py
```diff
@@ -50,6 +50,7 @@
     UnischemaField('matrix_nullable', np.uint16, _DEFAULT_IMAGE_SIZE, NdarrayCodec(), True),
     UnischemaField('sensor_name', np.unicode_, (1,), NdarrayCodec(), False),
     UnischemaField('string_array_nullable', np.unicode_, (None,), NdarrayCodec(), True),
+    UnischemaField('integer_nullable', np.int32, (), ScalarCodec(ShortType()), True),
```
SashaOv: One more reason to consider more "organized" schema where you can set flags on fields :-)

SashaOv: Any way we can set it as a flar (like nullable=True) in the future API?

petastorm/py_dict_reader_worker.py
```diff
@@ -226,11 +226,13 @@ def _load_rows_with_predicate(self, pq_file, piece, worker_predicate, shuffle_ro
             return filtered_decoded_predicate_rows
 
     def _read_with_shuffle_row_drop(self, piece, pq_file, column_names, shuffle_row_drop_partition):
+        # If integer_object_nulls is set to False, nullable integer fields are return as floats
+        # with nulls translated to nans
```
SashaOv: Did you also update the doc of `to_pandas` ?

petastorm/tests/test_common.py
selitvin: We can do it today. Chose not to do since it would be too verbose.

petastorm/py_dict_reader_worker.py
selitvin: `to_pandas` is part of pyarrow.

selitvin: `to_pandas` is part of pyarrow.

PR #340, opened 2019-03-28
"Fix unqualified exec is not allowed in function" SyntaxError

selitvin:
For some reason, I am seing this issue locally, while tests we passing on Travis.

To mitigate, moved new `namedtuple_gt_255_fields` out of the codeblock that is visible to Python 2.7 compiler.

@remysaissy, can you please take a look?


remysaissy:
Since the namedtuple.py's purpose is for the namedtuple_255 implementation, maybe we could rather put the six.PY2 condition in the tools.py module and make namedtuple.py out of Python2. That would save one new file.

Otherwise looks good.

PR #344, opened 2019-04-02
Add `Reader.reset()` method.

selitvin:
Resets ``Reader`` state and allows to fetch more samples once the ``Reader`` finished reading all epochs,
as specified by the ``num_epochs`` parameter.

Once all samples were read from a reader, an attempt to fetch new sample (e.g. ``next(reader)`` would raise
``StopIterationError``. You can reset the reader to the original state and restart reading samples
calling ``reset()``.

We do not support calling ``reset()`` until all samples were consumed. ``NotImplementedError``
will be raised if a user attempt to do so.

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -454,17 +454,43 @@ class Reader(object):
         # 4. Create a rowgroup ventilator object
         normalized_shuffle_row_drop_partitions = \
             self._normalize_shuffle_options(shuffle_row_drop_partitions, self.dataset)
-        ventilator = self._create_ventilator(filtered_row_group_indexes, shuffle_row_groups,
-                                             normalized_shuffle_row_drop_partitions, num_epochs, worker_predicate,
-                                             self._workers_pool.workers_count + _VENTILATE_EXTRA_ROWGROUPS)
+        self.ventilator = self._create_ventilator(filtered_row_group_indexes, shuffle_row_groups,
+                                                  normalized_shuffle_row_drop_partitions, num_epochs, worker_predicate,
+                                                  self._workers_pool.workers_count + _VENTILATE_EXTRA_ROWGROUPS)
 
         # 5. Start workers pool
         self._workers_pool.start(worker_class, (pyarrow_filesystem, dataset_path, storage_schema, self.ngram,
                                                 row_groups, cache, transform_spec),
-                                 ventilator=ventilator)
+                                 ventilator=self.ventilator)
         logger.debug('Workers pool started')
 
         self.last_row_consumed = False
+
+    def reset(self):
+        """Resets ``Reader`` state and allows to fetch more samples once the ``Reader`` finished reading all epochs,
+        as specified by the ``num_epochs`` parameter.
+
+        Once all samples were read from a reader, an attempt to fetch new sample (e.g. ``next(reader)`` would raise
+        ``StopIterationError``. You can reset the reader to the original state and restart reading samples
+        calling ``reset()``.
+
+        We do not support calling ``reset()`` until all samples were consumed. ``NotImplementedError``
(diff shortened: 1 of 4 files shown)


petastorm/workers_pool/ventilator.py
```diff
@@ -120,6 +122,10 @@ def completed(self):
         assert self._iterations_remaining is None or self._iterations_remaining >= 0
         return self._stop_requested or self._iterations_remaining == 0 or not self._items_to_ventilate
 
+    def reset(self):
+        self._iterations_remaining = self._iterations
```
rgruener: Wouldnt you need to stop the currently running thread first before resetting? I guess since you require the Reader to be completed before resetting the thread would be done anyways though there are no protections in this class from calling reset while still ventilating which would result in multiple threads adding to the ventilation queue.  Perhaps not a problem now but could develop into a bug later.

petastorm/reader.py
```diff
@@ -454,18 +454,42 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
...
+        Once all samples were read from a reader, an attempt to fetch new sample (e.g. ``next(reader)`` would raise
+        ``StopIterationError``. You can reset the reader to the original state and restart reading samples
+        calling ``reset()``.
+
+        We do not support calling ``reset()`` until all samples were consumed. ``NotImplementedError``
+        will be raised if a user attempt to do so.
+
+        :return: None
+        """
+        # TODO(yevgeni): could there be a race here?
```
rgruener: Thinking about it I dont think there should be a race worry here if you are waiting for reader to be completed but obviously I cant be 100% sure

petastorm/reader.py
```diff
@@ -454,18 +454,42 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
...
 
+    def reset(self):
+        """Resets ``Reader`` state and allows to fetch more samples once the ``Reader`` finished reading all epochs,
+        as specified by the ``num_epochs`` parameter.
+
+        Once all samples were read from a reader, an attempt to fetch new sample (e.g. ``next(reader)`` would raise
+        ``StopIterationError``. You can reset the reader to the original state and restart reading samples
+        calling ``reset()``.
+
+        We do not support calling ``reset()`` until all samples were consumed. ``NotImplementedError``
```
rgruener: Might be worth mentioning you cannot call reset once a Reader has been joined/stopped (since that might not be clear)

petastorm/reader.py
```diff
@@ -454,18 +454,42 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
...
+        will be raised if a user attempt to do so.
+
+        :return: None
+        """
+        # TODO(yevgeni): could there be a race here?
+        if not self.last_row_consumed:
+            # Don't allow reseting in the middle of epoch iterations since it is not very well defined how
```
rgruener: I feel like if somebody wanted to reset in the middle of an iteration it would be fine to drop the in flight samples (as that is the behavior I would expect). Though I can't think of a reason somebody would need that so I think its fine to not implement that for now (especially since there are likely race conditions with that one)

selitvin: Agree. My train of thought was that it would be hard to guarantee anything (either we drop, or we don't, or drop some?). A clean way is probably to drop existing workers and existing output queue and replace them with new ones (that would make sure we kill all inflights). Wanted to avoid this kind of undefined behavior.

petastorm/workers_pool/ventilator.py
selitvin: Good point. Added protection. Also add tests to cover ventilator behavior better.

PR #345, opened 2019-04-03
Rename `Unischema.name` to `Unischema._name` property to avoid conflict user fields named `name`.

selitvin:
"name" is a likely name for a user column. Existing `Unishema.name` attribute conflicts with
dynamically generated attributes. We rename `Unischema.name` into `Unischema._name`. If there
is a conflict between user field and existing attribute, we emit a warning and skip dynamic
attribute creation for that field.

This is a second attempt for a fix. Previous one broke legacy dataset reading and was reverted.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -134,7 +134,11 @@ class Unischema(object):
         self._fields = OrderedDict([(f.name, f) for f in sorted(fields, key=lambda t: t.name)])
         # Generates attributes named by the field names as an access syntax sugar.
         for f in fields:
-            setattr(self, f.name, f)
+            if not hasattr(self, f.name):
+                setattr(self, f.name, f)
+            else:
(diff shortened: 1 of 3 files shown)


petastorm/unischema.py
```diff
@@ -197,10 +201,6 @@ def __str__(self):
     def fields(self):
         return self._fields
 
-    @property
```
Ivan-Dimitrov: Shouldn't you add a property _name

selitvin: I was trying to do that before, so I changed the name of the `_name` attribute to `__name`. That resulted in inability to depickle the unischema in existing datasets. It was a mistake to use pickling to store Unischema as pickle from the start. It makes it much harder to change the code without breaking backward compatibility. For now, leaving _name an attribute (which is still accessible by users if really needed). The only downside is that a user may assign to `_name`. I would think this is not a very big deal though...

Issue #332, opened 2019-03-14
Make it easier to specify fields in n-gram


selitvin:
`make_reader` takes either a list of UnischemaField objects or regular expressions. Using regular expression in `schema_fields=` argument simplifies user code. `NGram` takes only `UnischemaFields` as a parameter. 

Let's follow the same pattern and allow user to pass a list of regular expressions during `NGram` construction.

jpugliesi:
@selitvin If i'm not mistaken, looks like this can be closed since #365 is complete

selitvin:
Oh, yes, thank you!

PR #365, opened 2019-05-02
adding regex expressions for ngrams

Ivan-Dimitrov:
#332 

Letting ngram take regular expressions for it's field and timestamp_field arguments. 

The regex expressions are converted to unischema fields when the ngram is initialized, so it shouldn't have downstream impacts.

However, it does add an argument (schema) to NGram, so users will have to update their usage.

@selitvin Some of the method functions for ngram take schema as an argument, I think it would be best to remove that and just have schema be a class parameter.  Do you see a reason not to do this?


petastorm/ngram.py
```diff
@@ -98,13 +99,14 @@ class NGram(object):
     key will be the item.
     """
 
-    def __init__(self, fields, delta_threshold, timestamp_field, timestamp_overlap=True):
+    def __init__(self, fields, delta_threshold, timestamp_field, schema, timestamp_overlap=True):
```
selitvin: The thing is that when we create NGram instance we don't yet have the schema (unless we load it manually from the dataset). This is inconvenient.

E.g.

```python
ngram=NGram(...)
with make_reader(url, schema_fields=ngram) as reader:
   ...
```
Then a user would have to do:
```python
schema=load_schema_from_dataset(url) # don't remember the exact function.
ngram=NGram(...)
with make_reader(url, schema_fields=ngram) as reader:
   ...
```

One way to address this is to defer resolution of regex to later, i.e.  

```python
class NGram(object):
...
   def resolve_regex(schema):
      ....
```

then in `make_reader`, once we have the schema at hand we call `ngram.resolve_regex(schema)`.

What do you think?

That way we also don't have to break compatibility with the existing code.

Ivan-Dimitrov: I like the option for make_reader to call resolve_regex. I will change it to that.

Ivan-Dimitrov: A follow up, considering that users have to give fields/timestamp_field with a unischema field object.  I might be wrong, but most are probably loading in the schema before creating an ngram.

selitvin: What users would do today, they would take an instance of the schema from the code (and not the dataset). This is not great because they could be out of sync. Maybe we just allow it to be a string?

Ivan-Dimitrov:
I'm using pylint locally but it doesn't seem to catch all the lint issues. What linter is used for this repo?

petastorm/tests/test_ngram_end_to_end.py
```diff
@@ -539,3 +539,75 @@ def test_ngram_shuffle_drop_ratio(synthetic_dataset, reader_factory):
...
+    expected_fields = [TestSchema.id, TestSchema.id2, TestSchema.id_float, TestSchema.id_odd,
+                       TestSchema.sensor_name, TestSchema.partition_key]
+
+    ngram = _test_continuous_ngram_returns(fields, ts_field, dataset_num_files_1, reader_factory)
+
+    # fields should get resolved after call to a reader
+    ngram_fields = ngram.fields
+    for exp_field in expected_fields:
+        for k in ngram_fields.keys():
+            assert exp_field in ngram_fields[k]
```
selitvin: If ngram fields is a superset of `expected_fields`, I think this check would not cover that. Consider stronger comparison, something like:
```python
assert {'id', 'id2', ...} == set(ngram_fields.keys())
```

petastorm/tests/test_ngram_end_to_end.py
```diff
@@ -539,3 +539,75 @@ def test_ngram_shuffle_drop_ratio(synthetic_dataset, reader_factory):
...
+    expected_fields = [TestSchema.id, TestSchema.id2, TestSchema.id_float, TestSchema.id_odd,
+                       TestSchema.sensor_name, TestSchema.partition_key]
+
+    ngram.resolve_regex_field_names(TestSchema)
+
+    ngram_fields = ngram.fields
+
+    for exp_field in expected_fields:
+        for k in ngram_fields.keys():
+            assert exp_field in ngram_fields[k]
```
selitvin: Same here. This is a weaker check then we could easily have.

petastorm/ngram.py
```diff
@@ -187,6 +192,12 @@ def _ngram_pass_threshold(self, ngram):
                 return False
         return True
 
+    def resolve_regex_field_names(self, schema):
+        """Resolve string(s) (regular expression(s)) that were passed into 'fields' and 'timestamp_field' parameters
+        """
+        self._fields = {k: self.convert_fields(schema, self._fields[k]) for k in self._fields.keys()}
+        self._timestamp_field = self.convert_fields(schema, [self._timestamp_field])[0]
```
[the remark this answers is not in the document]
Ivan-Dimitrov: Good point

petastorm/ngram.py
```diff
@@ -145,30 +148,32 @@ def delta_threshold(self):
...
 
         for key in fields:
             if not isinstance(fields[key], list):
-                raise ValueError('Each field value must be a list of unischema fields')
+                raise ValueError('Each field value must be a list of unischema fields/string (regular expressions)')
             for field in fields[key]:
-                if not isinstance(field, UnischemaField):
-                    raise ValueError('All field values must be of type UnischemaField.')
+                if not (isinstance(field, UnischemaField) or isinstance(field, string_types)):
+                    raise ValueError('All field values must be of type UnischemaFieldstring (regular expressions)')
```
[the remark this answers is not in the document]
Ivan-Dimitrov: missing `/`

petastorm/tests/test_ngram_end_to_end.py
Ivan-Dimitrov: changed

petastorm/tests/test_ngram_end_to_end.py
Ivan-Dimitrov: changed

PR #382, opened 2019-06-04
Support inter-row-group shuffling queue when reading from pytorch

selitvin:
Added `shuffling_queue_capacity` and `min_after_dequeue` arguments to `petastorm.pytorch.DataLoader` class. When these values are set, an inmemory buffer is used to accumulate and reshuffle data being loaded. 

The mechanism is not as flexible as PyTorch's sampling which is hard to implement because of the to have an efficient random data access in Parquet files.

### DIFF
--- a/petastorm/pytorch.py
+++ b/petastorm/pytorch.py
@@ -12,17 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Must import pyarrow before torch. See: https://github.com/uber/petastorm/blob/master/docs/troubleshoot.rst
-import re
-
-import pyarrow  # noqa: F401 pylint: disable=W0611
-
 import collections
 import decimal
+# Must import pyarrow before torch. See: https://github.com/uber/petastorm/blob/master/docs/troubleshoot.rst
+import re
 
 import numpy as np
 from six import PY2
 from torch.utils.data.dataloader import default_collate
+
+from petastorm.reader_impl.shuffling_buffer import RandomShufflingBuffer, NoopShufflingBuffer
 
 if PY2:
     _string_classes = basestring  # noqa: F821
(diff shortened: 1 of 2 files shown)


petastorm/pytorch.py
```diff
@@ -97,7 +96,8 @@ class DataLoader(object):
     runs out of samples.
     """
 
-    def __init__(self, reader, batch_size=1, collate_fn=decimal_friendly_collate):
+    def __init__(self, reader, batch_size=1, collate_fn=decimal_friendly_collate,
+                 shuffling_queue_capacity=0, min_after_dequeue=0):
```
Ivan-Dimitrov: Add param descriptions for the new arguments

petastorm/pytorch.py
```diff
@@ -120,11 +125,39 @@ def __iter__(self):
...
+        self._shuffling_buffer.finish()
+
+        # Some code duplication with the main reading loop. Not sure how to reorganize to avoid this without
+        # overcomplicating the code too much.
+        while self._shuffling_buffer.can_retrieve():
+            post_shuffled_row = self._shuffling_buffer.retrieve()
+            batch.append(post_shuffled_row)
+
+            # Batch is ready? Collate and emmit
             if len(batch) == self.batch_size:
```
Ivan-Dimitrov: nit: I think the final 'if batch` check is fine. But can't we just have this be len(batch) <= self.batch_size. 
Is there any situation where len(batch) would be > self.batch_size here.

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -127,3 +126,26 @@ def test_decimal_friendly_collate_input_has_decimals_in_tuple():
...
     assert desired[0] == actual[0]
     np.testing.assert_equal(desired[1], actual[1].numpy())
+
+
+@pytest.mark.parametrize('reader_factory', MINIMAL_READER_FLAVOR_FACTORIES)
+def test_no_shuffling(synthetic_dataset, reader_factory):
+    with DataLoader(reader_factory(synthetic_dataset.url, schema_fields=['^id$'], workers_count=1,
+                                   shuffle_row_groups=False)) as loader:
+        ids = [row['id'][0].numpy() for row in loader]
+        np.testing.assert_array_equal(range(len(ids)), ids)
```
Ivan-Dimitrov: I don't really understand this check. Should we be check the len(ids) vs the len(synthetic_dataset)?

selitvin: You are right. Made it clearer.

petastorm/pytorch.py
selitvin: Done. Added more documentation

petastorm/tests/test_pytorch_dataloader.py
Ivan-Dimitrov: Okay, the tests are clearer now.

petastorm/pytorch.py
selitvin: Not sure I follow. The second `while self._shuffling_buffer.can_retrieve()` loop on line 168, runs down the sample while completely depleting the shuffling queue. 
For example:
- Assume batch size is 9
- In the stead state the shuffling queue has 1000 samples.
- Now we don't have any more samples to load and `for row in self.reader:` exits. We still have 1000 samples awaiting in the shuffling queue.
- We call finish, which enables further depletion of the samples
- We are still able to produce 111 batches of size 9, then pop last sample from the shuffling queue and publish it on line 178.

Did I miss a way to simplify something?

Ivan-Dimitrov: Yes, that makes sense. My initial thought that after row in self.reader exists, you can at most have one batch left. That was incorrect. 

Thanks for explaining

PR #383, opened 2019-06-05
Support reading from a partitioned dataset. Interpret types of the partition-by scalars properly. Also, remove dependency on pyspark while reading using make_batch_reader.

selitvin:
We use pyarrow resolved types for the partition-by key (in `unischema.py`) and propagate these types to the proper configuration of UnischemaField.
    
This diff also removes `import pyspark` statements from the `make_petatsorm_reader` import/execution path to make sure it can be used without having pyspark installed in the python environment.

### DIFF
--- a/petastorm/tests/test_unischema.py
+++ b/petastorm/tests/test_unischema.py
@@ -19,6 +19,7 @@ from decimal import Decimal
 
 import numpy as np
 import pyarrow as pa
+import pyarrow.parquet as pq
 import pytest
 from pyspark import Row
 from pyspark.sql.types import StringType, IntegerType, DecimalType, ShortType, LongType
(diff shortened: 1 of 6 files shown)


petastorm/tests/test_unischema.py
```diff
@@ -33,19 +34,19 @@
     from mock import mock
 
 
-def _mock_parquet_dataset(partition_names, schema):
+def _mock_parquet_dataset(partitions, schema):
     """Creates a pyarrow.ParquetDataset mock capable of returning:
 
         parquet_dataset.pieces[0].get_metadata(parquet_dataset.fs.open).schema.to_arrow_schema() == schema
-        parquet_dataset.partition_names.partition_names = partition_names
+        parquet_dataset.partitions = partitions (expected to be a list of pa.parquet.PartitionSet)
```
Ivan-Dimitrov: Shouldn't the parenthesis be a comment, not inline

PR #384, opened 2019-06-05
Adding the user parameter when pyarrow.hdfs.connect is called.

Ivan-Dimitrov:
User is added as optional parameters for filesystem resolver and various entry points
for file system resolver (materialize dataset, row group indexer, etc).

Updating unit tests

### DIFF
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -94,7 +94,7 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_
     spark_config = {}
     _init_spark(spark, spark_config, row_group_size_mb, use_summary_metadata)
     yield
-
+    print ("testing")
     # After job completes, add the unischema metadata and check for the metadata summary file
     if filesystem_factory is None:
         resolver = FilesystemResolver(dataset_url, spark.sparkContext._jsc.hadoopConfiguration(), user=spark.sparkContext.sparkUser())
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -50,7 +50,7 @@ class PetastormMetadataGenerationError(Exception):
 
 @contextmanager
 def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_summary_metadata=False,
-                        filesystem_factory=None, user=None):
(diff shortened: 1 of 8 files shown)


Ivan-Dimitrov:
Two decisions on this loom.

(1) I didn't add a user argument when people are only reading data/metadata. Usually the read permissions of the files aren't limited, but there is no reason this should be true across all use cases.

(2) Are we sure we want to change all the files I have changed. It's is an optional argument with None as a default, so we should break any user code.

Ivan-Dimitrov:
The five locations where we only read with filesystem resolver are `petastorm/reader/make_reader`, `petastorm/etl/metadata_util`, `get_schema_from_dataset_url` in `petastorm/etl/dataset_metadata`, `petastorm/reader_impl/row_group_loader`, and `petastorm/spark_utils`. 

I am starting to lean towards adding the user param. Behavior won't change since it will remain "None" unless changed. But I'm not sure what the standard should be. 

I have reviewed all the files I did change, and the user param is needed since they all write.

petastorm/etl/dataset_metadata.py
```diff
@@ -89,15 +89,15 @@ def materialize_dataset(spark, dataset_url, schema, row_group_size_mb=None, use_
       indexing method. The custom indexing method is more scalable for very large datasets.
     :param pyarrow_filesystem: A pyarrow filesystem object to be used when saving Petastorm specific metadata to the
       Parquet store.
-
+    :param user: String denoting username when connecting to HDFS
```
selitvin: Maybe mention a word about the default behavior (defined by the driver used)

petastorm/hdfs/namenode.py
```diff
@@ -2,7 +2,7 @@
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
+# You may obtain a copy of the License a
```
selitvin: unintended change

petastorm/etl/dataset_metadata.py
Ivan-Dimitrov: done

petastorm/hdfs/namenode.py
Ivan-Dimitrov: fixed

PR #393, opened 2019-06-17
Make sure in_pseudorandom_split works also with non-string fields.

selitvin:
We convert all values to string and then calculate the hash as the value was a string.

### DIFF
--- a/petastorm/tests/test_predicates.py
+++ b/petastorm/tests/test_predicates.py
@@ -107,14 +107,28 @@ def test_or_argegarion(all_values):
             assert included_values == values1.union(values2)
 
 
-def test_pseudorandom_split(all_values):
+def test_pseudorandom_split_on_string_field(all_values):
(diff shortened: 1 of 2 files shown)


petastorm/tests/test_predicates.py
```diff
@@ -107,14 +107,28 @@ def test_or_argegarion(all_values):
...
+                included_values.add(val)
+        expected_num = values_num * split_list[idx]
+        assert pytest.approx(len(included_values), expected_num * 0.1) == expected_num
+
+
+def test_pseudorandom_split_on_integer_field():
+    split_list = [0.3, 0.4, 0.1, 0.0, 0.2]
+    int_values = list(range(1000))
+    values_num = len(int_values)
+    for idx in range(len(split_list)):
```
jsgoller1: Minor nit, but I think that `for idx, _ in enumerate(split_list)` is slightly easier to read.

selitvin: That's a nice idea. Done.

Issue #391, opened 2019-06-14
Train-Test Dataset Split


seranotannason:
Is there currently support for splitting a Petastorm dataset into train-test for PyTorch? In PyTorch, one would typically do this to a Dataset class but since Petastorm only has the classes Reader and DataLoader (as below), I wonder if this feature has been implemented.

`trainloader = DataLoader(make_reader('file://' + filename), batch_size=128)`

selitvin:
You can use `make_reader(..., predicate=in_lambda(['sample_id'], lambda x: x % 10 < 2))`. The predicate will be called for each sample (or batch in case of `make_batch_reader`). You can decide which rows are included in the current split. 

We have off-the-shelf `in_pseudorandom_split` predicate which implements this kind of partitioning for you. For example:
`make_reader(..., predicate=in_pseudorandom_split([0.8, 0.2], 0, 'sample_id'))` will return 80% of the samples you can use a training set. Using `in_pseudorandom_split([0.8, 0,2], 1, 'sample_id')` would return the complimenting 80%. 

Note that #393 makes the function work also with non-string fields.

In order for this approach to work more effectively and avoid loading and (automatically) discarding data from the split you are not interested in, it would be recommended to keep your parquet sorted or partitioned by the field you use for your split criteria.

seranotannason:
Consider the following block when the schema has the columns `filename`, `image`, and `label`:

```
def _transform_row_train(cifar_row):
    result_row = {
        'image': cifar_row['image'],
        'label': cifar_row['label']
    }
    return result_row

transform_spec_train = TransformSpec(_transform_row_train, removed_fields=['filename'])

trainloader = DataLoader(make_reader(filename, 
    predicate=in_pseudorandom_split([0.75, 0.25], 0, 'filename'), 
    transform_spec=transform_spec_train))
```

This block will produce an error that cifar_row has no keys 'image' or 'label'. It turns out that `transform_spec` performs its operations only on the columns specified as the third argument of `predicate`. Is this the desired functionality?

selitvin:
Nope. That was not an aware choice. I think you are right. It makes sense to filter first by predicate and then apply the transform. Actually, with `make_batch_reader` the order is correct. We should fix it. 

Created #394 with a fix. Can you please try and let me know if it works for you?

seranotannason:
Both #393 and #394 work!

selitvin:
These are good news! Will include them in the next release probably within couple of days up to two weeks from now.

[...]

PR #394, opened 2019-06-18
Apply transform function after predicate was evaluated (with make_reader)

selitvin:
It was reported in #391, that having the transform applied before predicate
does not allow us to get rid of string fields that can not be forwarded to PyTorch.

This PR changes the order of predicate/transform so that the scenario described in #391 works.

### DIFF
--- a/petastorm/py_dict_reader_worker.py
+++ b/petastorm/py_dict_reader_worker.py
@@ -192,9 +192,8 @@ class PyDictReaderWorker(WorkerBase):
                                                           shuffle_row_drop_partition)
 
         # Decode values
-        transform_func = self._transform_spec.func if self._transform_spec else (lambda x: x)
         decoded_predicate_rows = [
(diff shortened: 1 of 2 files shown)


petastorm/py_dict_reader_worker.py
```diff
@@ -221,9 +220,14 @@ def _load_rows_with_predicate(self, pq_file, piece, worker_predicate, shuffle_ro
...
-            return all_cols
+            result = all_cols
         else:
-            return filtered_decoded_predicate_rows
+            result = filtered_decoded_predicate_rows
+
+        if self._transform_spec.func:
+            result = [self._transform_spec.func(row) for row in result]
+
+            return result
```
Ivan-Dimitrov: Should this be indented? Are we returning anything here when there isn't a transform function?

jsgoller1: :+1:  to this

selitvin: Nice catch! The tests were also failing because of that.

PR #407, opened 2019-07-02
Cast int to string in Runtime error (ArrowReaderWorker)

praateekmahajan:
In ArrowReaderWorker, when trying to load a column of list type, where list is of variable length an error needs to be thrown. This error is defined here.

However the error fails while printing

Instead of
`', '.join({values.shape[0] for value in ....})`

It should be
`', '.join({str(values.shape[0]) for value in ....})`

### DIFF
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -61,9 +61,10 @@ class ArrowReaderWorkerResultsQueueReader(object):
                     try:
                         result_dict[column.name] = np.vstack(list_of_lists.tolist())
                     except ValueError:
-                        raise RuntimeError("""Length of all values in column '%s' are expected to be the same length.
-                                           Got the following set of lengths: '%s'""" %
-                                           (column.name, ', '.join({str(value.shape[0]) for value in list_of_lists})))
+                        raise RuntimeError('Length of all values in column \'{}\' are expected to be the same length. ' 
+                                           'Got the following set of lengths: {}'
+                                           .format(column.name,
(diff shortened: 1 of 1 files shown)


petastorm/arrow_reader_worker.py
```diff
@@ -61,10 +61,9 @@ def read_next(self, workers_pool, schema, ngram):
                     try:
                         result_dict[column.name] = np.vstack(list_of_lists.tolist())
                     except ValueError:
-                        raise RuntimeError('Length of all values in column \'%s\' are expected to be the same length. '
-                                           'Got the following set of lengths: \'%s\'',
-                                           column.name,
-                                           ', '.join({value.shape[0] for value in list_of_lists}))
+                        raise RuntimeError("""Length of all values in column '%s' are expected to be the same length.
```
selitvin: Thank you for the fix!
To better align with coding style in the rest of the codebase, can you please:
1. Implement multiline string and string formating like this:
```python
('Length of all values {}...'
 'Got the folling {} ...')
.format(column.name, ', '.join(str(value.shape[0]) for value in list_of_lists)))
```
2. Do you need to construct a set (the curly brackets)? I think an iterable should work fine, i.e.
```python
', '.join({str(value.shape[0]) for value in list_of_lists})
```
-->
```python
', '.join(str(value.shape[0]) for value in list_of_lists)
```

praateekmahajan: Thank you for your comment. 
Added multiline string formatting and removed the set :)

selitvin:
Oops, minor lint error . Updated the PR for you. Will land it once the tests passes.

PR #411, opened 2019-07-31
Read proper batches when using petastorm.pytorch.DataLoader with make…

selitvin:
…_batch_reader.

Before this fix, batches were not formed properly. To proper support make_batch_rader with DataLoader we have:
- Added an attribute to a Reader object instance indicating that it returns batched data (these are Parquet row groups)
- For petastorm.pytorch.DataLoader that work with a batched reader, we store the entire batch into a shuffling buffer.


praateekmahajan:
Sorry, for spamming the PR, but I came across the [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) released by PyTorch which might be helpful..

selitvin:
> Sorry, for spamming the PR, but I came across the [IterableDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.IterableDataset) released by PyTorch which might be helpful..
@praateekmahajan , thanks for pasting the link. Very useful. Did not see that one yet.

petastorm/pytorch.py
```diff
@@ -122,60 +122,84 @@ def __init__(self, reader, batch_size=1, collate_fn=decimal_friendly_collate,
...
-                    yield self.collate_fn(batch)
-                    batch = []
-
-        # Once reader can not emit new rows, we might still have a bunch of rows waiting in the shuffling buffer.
-        # Telling shuffling buffer thatt we are finished allows to deplete the buffer completely, regardless its
+            if not self.reader.is_batched_reader:
+                self._shuffling_buffer.add_many([row_as_dict])
+            else:
+                # Transposition:
+                #   row_as_dict:        {'a': [1,2,3], 'b':[4,5,6]}
```
Ivan-Dimitrov: Is it always true that a and b will have the same length of data

selitvin: Good point. It should be so for scalar fields. Need to check if it works properly in case of lists.

PR #412, opened 2019-08-01
Fix pyspark hello world link

mtn:
Just a broken link.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -254,7 +254,7 @@ SQL can be used to query a Petastorm dataset:
       'SELECT count(id) '
       'from parquet.`file:///tmp/hello_world_dataset`').collect()
 
-You can find a full code sample here: `pyspark_hello_world.py <examples/hello_world/petastorm_dataset/pyspark_hello_world.py>`_,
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -254,7 +254,7 @@ SQL can be used to query a Petastorm dataset:
       'SELECT count(id) '
       'from parquet.`file:///tmp/hello_world_dataset`').collect()
 
-You can find a full code sample here: `pyspark_hello_world.py <https://github.com/uber/petastorm/blob/master/examples/hello_world/pyspark_hello_world.py>`_,
+You can find a full code sample here: `pyspark_hello_world.py <examples/hello_world/petastorm_dataset/pyspark_hello_world.py>`_,
```
selitvin: Thank you for your contribution!

Can we please change it to be a full url: `https://github.com/uber/petastorm/blob/master/examples/hello_world/petastorm_dataset/pyspark_hello_world.py`? The reason for using full url is that otherwise, the link would be broken in the autogenerated documentation here: https://petastorm.readthedocs.io/en/latest/readme_include.html#id8

PR #414, opened 2019-08-09
Use deque in NoopShufflingBuffer.

selitvin:
Previous implementation used list. pop(0) from a list is very expensive.

Timing of the new code shows x25 speedup.

### DIFF
--- a/petastorm/reader_impl/shuffling_buffer.py
+++ b/petastorm/reader_impl/shuffling_buffer.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import abc
+from collections import deque
 
 import numpy as np
 import six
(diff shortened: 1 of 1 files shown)


selitvin:
Used this small snippet to test the performance:
```python
import time

from petastorm.reader_impl.shuffling_buffer import NoopShufflingBuffer, RandomShufflingBuffer


def foo():
    # buf = RandomShufflingBuffer(1000000, 1, extra_capacity=1000)
    buf = NoopShufflingBuffer()
    for _ in range(1):
        buf.add_many([0] * 100000)
    i = 0
    while buf.can_retrieve():
        i = i + 1
        buf.retrieve()

    print(i)
t=time.time()
foo()
print(time.time() - t)
```
With list: 1.25 sec
With deque: 0.047 sec

Issue #416, opened 2019-08-13
removed_fileds not acknowledged in arrow_reader_worker


praateekmahajan:
Sadly won't be able to reproduce the error quickly :

```python
transform = TransformSpec(lambda x : x, removed_fields=["columnA"])
reader = make_batch_reader(dataset_url, transform_spec=transform)
# next(reader) gives an error
```

I was trying to trace the error and realised that 
1. `reader.schema` : doesn't have `columnA`
2. `arrow_reader_worker.py` calls `schema.make_namedtuple(**result_dict)` where `result_dict` comes from worker `workers_pool.get_results()`. 
The `result_dict` seems to have `columnA` in it.

The error is following :

```python
~/Desktop/github/petastorm/petastorm/reader.py in __next__(self)
    598     def __next__(self):
    599         try:
--> 600             return self._results_queue_reader.read_next(self._workers_pool, self.schema, self.ngram)
    601         except StopIteration:
    602             self.last_row_consumed = True

~/Desktop/github/petastorm/petastorm/arrow_reader_worker.py in read_next(self, workers_pool, schema, ngram)
     71             print(f"The schema has the following keys : \n {schema._fields.keys()}")
     72             print(f"The resulting dict has the folowing keys {result_dict.keys()}")
---> 73             return schema.make_namedtuple(**result_dict)
     74 
     75         except EmptyResultError:

~/Desktop/github/petastorm/petastorm/unischema.py in make_namedtuple(self, **kargs)
    233                 typed_dict[key] = None
    234         print(typed_dict.keys())
--> 235         return self._get_namedtuple()(**typed_dict)
    236 
    237     def make_namedtuple_tf(self, *args, **kargs):

TypeError: __new__() got an unexpected keyword argument 'columnA'
```

praateekmahajan:
Suggested (quick) solution is to add an if condition inside the `read_next` to ensure that `column.name in schema._field.keys()`.

One can have a look here at my fork [here](https://github.com/praateekmahajan/petastorm/commit/40ba510e2325847f63e195cd7919a4d93eb617fa).

selitvin:
Seems to me that the problem her is what appears to be a confusing interface. I see that you assumed that 
```python
transform = TransformSpec(lambda x : x, removed_fields=["columnA"])
```
removes `"columnA"` from the data-frame for you, while my intent was to indicate that your lambda actually removes the column. I can see how this is confusing. I would expect the code to be like this:
```python
def delColumnA(x):
[...]

PR #417, opened 2019-08-13
Automatically delete columns when TransformSpec(..., removed_fields=.…

selitvin:
….) is specified.

Previously, we were relying on a user to delete fields inside TransformSpec's func implementation:

```python
def delColumnA(x):
   del x['columnA']
   return x

transform = TransformSpec(delColumnA, removed_fields=["columnA"])
```

This assumption is non-trivial and induce additional burden on a user (see #416). A simpler way is to do the column removal for a user, so the following code would be equivalent to the previous snippet we showedL
```python
transform = TransformSpec(removed_fields=["columnA"])
```


petastorm/py_dict_reader_worker.py
```diff
@@ -34,6 +34,24 @@ def _merge_two_dicts(a, b):
...
+        for row in all_rows:
+            for field_to_remove in transform_spec.removed_fields:
+                if field_to_remove in row:
+                    del row[field_to_remove]
+
+    return all_rows
+
+
 def _select_cols(a_dict, keys):
     """Filteres out entries in a dictionary that have a key which is not part of 'keys' argument. `a_dict` is not
```
praateekmahajan:
```suggestion
    """Filters out entries in a dictionary that have a key which is not part of 'keys' argument. `a_dict` is not
```

While we are at this file, possibly a good idea to get rid of the typo! 🔍

petastorm/py_dict_reader_worker.py
```diff
@@ -34,6 +34,24 @@ def _merge_two_dicts(a, b):
...
+    """Applies transform_spec.func to all rows in all_rows (if func is specified). Removes all fields as specified by
+    transform_spec.removed_fields, unless the fields were already deleted by transform_spec.func.
+
+    All modifications are performed in-place.
+    """
+    if transform_spec.func:
+        all_rows = [transform_spec.func(row) for row in all_rows]
+
+    if transform_spec.removed_fields:
+        for row in all_rows:
```
praateekmahajan: I was actually wondering if `all_rows` have a method is a dict or have a method that can fetch all their columns? For eg if it has `all_rows.columns` or `all_rows.keys()`. If it does then we could just do something like

```python
for field_to_remove  in (all_rows.keys() & transform_spec.removed_fields):
    del row[field_to_remove]
```

petastorm/arrow_reader_worker.py
```diff
@@ -157,7 +157,21 @@ def _load_rows(self, pq_file, piece, shuffle_row_drop_range):
...
+            result_as_pandas = result.to_pandas()
+            # A user may omit `func` value if they intend just to delete some fields using the TransformSpec
+            if self._transform_spec.func:
+                transformed_result = self._transform_spec.func(result_as_pandas)
+            else:
+                transformed_result = result_as_pandas
+
+            # If transform function left a field that is listed in transform_spec's remove_fields, we remove it
+            # ourselves. Allows for the following transform-spec objects to be created:
+            # TransformSpec(removed_fields=['some field'])
```
praateekmahajan: Similar comment as in `py_dict_reader_worker` could save few lines here. (I believe same time complexity though)

```python
for field_to_remove  in (transformed_result.columns & transform_spec.removed_fields):
    del row[field_to_remove]
```

petastorm/transform.py
```diff
@@ -26,8 +26,9 @@ def __init__(self, func, edit_fields=None, removed_fields=None):
         ``edit_fields`` and ``removed_fields`` define mutating operations performed on the original schema that
         produce a post-transform schema. ``func`` return value must comply to this post-transform schema.
 
-        :param func: A callable. The function is called on the worker thread. It takes a dictionary that complies to
-          the input schema and must return a dictionary that complies to a post-transform schema.
+        :param func: Optional. A callable. The function is called on the worker thread. It takes a dictionary that
+          complies to the input schema and must return a dictionary that complies to a post-transform schema. User may
+          omit this argument and specify only `remove_fields` argument.
```
praateekmahajan:
```suggestion
          In case the user wants to only remove certain fields, the user may omit this argument and specify only `remove_fields` argument.
```

I believe this makes it slightly more clear.

PR #421, opened 2019-08-30
Make 'codec' and 'nullable' optional UnischemaField arguments

selitvin:
The goal is twofold:
- Spare users' effort of defining codec attribute for fields that are scalars and we can easily decode/encode the values ourselves.
- Make it possible for a user to skip the encoding/decoding mechanism (this is an experimental feature at the moment)

If no codec is specified, we try to deduce spark type from `UnischemaField`''s `numpy_dtype` attribute when possible. If `numpy_dtype` is not set or different from a numpy scalar or a `Decimal`, we just pass it through.

### DIFF
--- a/petastorm/tests/test_decode_row.py
+++ b/petastorm/tests/test_decode_row.py
@@ -11,11 +11,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from decimal import Decimal
 
 import numpy as np
 import pytest
+from pyspark.sql.types import DoubleType
 
-from petastorm.codecs import NdarrayCodec
+from petastorm.codecs import NdarrayCodec, ScalarCodec
 from petastorm.unischema import UnischemaField, Unischema
 from petastorm.utils import decode_row, DecodeFieldError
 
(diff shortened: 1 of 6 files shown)


petastorm/tests/test_decode_row.py
```diff
@@ -39,3 +41,50 @@ def test_can_not_decode():
 
     with pytest.raises(DecodeFieldError, match='matrix'):
         decode_row(row, MatrixSchema)
+
+
+def test_decode_numpy_scalar_when_codec_is_none():
+    """Decoding a row that has a field with the codec set to None. The type should be deduced automatically
```
rohankumar42: In the tests where the codec is being inferred, can we compare the codec inferred explicitly? For example, instantiate the expected `UnischemaField` and use the class's `__eq__`.

petastorm/unischema.py
```diff
@@ -34,11 +34,11 @@ def _fields_as_tuple(field):
     """Common representation of UnischemaField for equality and hash operators.
     Defined outside class because the method won't be accessible otherwise.
 
-    The only difference is that the type name of the ``codec`` field is returned
-    so that the codec object ID won't differentiate two otherwise identifcial
-    UniSchema fields.
+    Today codec instance also responsible for defining spark dataframe type. This knowledge should move
```
rohankumar42: :+1:

petastorm/unischema.py
```diff
@@ -110,8 +115,52 @@ def _new_gt_255_compatible_namedtuple(*args, **kwargs):
...
+    return getattr(_numpy_to_spark_mapping, cache_attr_name)
+
+
 # TODO: Changing fields in this class or the UnischemaField will break reading due to the schema being pickled next to
 # the dataset on disk
+def _field_spark_dtype(field):
+    if not field.codec:
+        if field.shape == ():
+            spark_type = _numpy_to_spark_mapping().get(field.numpy_dtype, None)
+            if not spark_type:
```
rohankumar42: Nit: 
```if spark_type is not None:```

petastorm/tests/test_decode_row.py
selitvin: Not sure what do you mean?

petastorm/unischema.py
selitvin: Done

petastorm/tests/test_decode_row.py
rohankumar42: Test `UnischemaField.codec` has the correct type

PR #424, opened 2019-09-04
Use UnischemaField object persisted in the dataset to guarantee correct decoding

selitvin:
Before this change, if a user was specifying `schema_fields=[Unischame('some_field', np.int8, (), NewCodec(), ...)]`
as a `make_reader` argument, we would use the new field definition, even that `some_field`'s `UnischemaField`
object persisted with the dataset had a different type/shape/codec.

This commit makes sure we use `some_field` field-name to lookup the corresponsing field name in the persisted dataset.
This guarantees correspondance between the field definition and actual persisted data.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -139,7 +139,7 @@ class Unischema(object):
 
         Fields can be either UnischemaField objects or regular expression patterns.
 
-        If one of the fields is not part of the schema an error is raised.
+        If one of the fields does not exist in this schema, an error is raised.
 
         The example returns a schema, with field_1 and any other field matching ``other.*$`` pattern.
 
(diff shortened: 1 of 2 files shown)


petastorm/unischema.py
```diff
@@ -152,24 +152,28 @@ def create_schema_view(self, fields):
...
-
-        for field in unischema_field_objects:
-            # Comparing by field names. Prevoiusly was looking for `field not in self._fields.values()`, but it breaks
-            # due to faulty pickling: T223683
-            if field.name not in self._fields:
-                raise ValueError('field {} does not belong to the schema {}'.format(field, self))
+        # For fields that are specified as instances of Unischema: make sure that this schema contains fields
+        # with these names.
+        exact_field_names = [field.name for field in unischema_field_objects]
+        unknown_field_names = set(exact_field_names) - set(self.fields.keys())
```
Ivan-Dimitrov: Why are we checking against self.fields.keys() here, shouldn't we be checking against the unischema

selitvin: We use only field names from what user gave to us and not UnischemaField instances to make 'new' codec field does not leak into the actual schema instances used for decoding. Otherwise, we could end up using new codec on data encoded with an old codec.

Ivan-Dimitrov: okay, I see

PR #428, opened 2019-09-17
allow NdarrayCodec.decode to pass through ndarrays

stedn:
We have a workflow where we had a dataset that was not encoded with NDArrayCodec, but we were decoding using it, essentially passing an ndarray through directly.  Was wondering if it is worthwhile to allow this behavior.  It's also possible there is another way to do this that we didn't detect.

### DIFF
--- a/petastorm/codecs.py
+++ b/petastorm/codecs.py
@@ -136,6 +136,8 @@ class NdarrayCodec(DataframeColumnCodec):
         return bytearray(memfile.getvalue())
 
     def decode(self, unischema_field, value):
+        if type(value) is np.ndarray:
+            return value
         memfile = BytesIO(value)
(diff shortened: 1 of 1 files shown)


stedn:
nevermind, that CLA thing is too onerous for 2 lines of code.

selitvin:
This is an interesting request. Can you please give a little bit more details about your setup? I am curious, are you reading from a parquet store? 

We saw an internal usecase where we use a storage format that is different from Parquet and that is capable of storing numpy arrays natively. We would need a similar functionality of by-passing the decoding. I wonder if your case is somewhat similar?   

I was working on #421 (in that PR codec=None can be set only for scalars for now). do you think specifying `codec=None` for ndarrasy would work for your case as well?

Issue #432, opened 2019-10-01
TF Dataset Runs Out Of Data Despite num_epochs=None


cupdike:
I have a petastorm dataset with 10000+ images.  I'm trying to train with 10 epochs and a batch size of 32.  Training runs are failing with:
`Your dataset ran out of data; interrupting training. Make sure that your dataset can generate at least `steps_per_epoch * epochs` batches (in this case, 3390 batches)` 

I am applying the `num_epochs=None` to make_reader which I thought was sufficient to supply "infinite" data:
```
with make_reader(dataset_url, reader_pool_type='dummy', num_epochs=None) as reader:

    dataset = make_petastorm_dataset(reader)
    dataset = dataset.map(lambda x: tf.numpy_function(func=process_inputs, inp=(x.image1, x.category), Tout=(tf.float32,tf.uint8)))

    dataset = dataset.batch(bs).shuffle(buffer_size=15000) 
```

 Am I doing something wrong here?

cupdike:
So it appears that there are different behaviors depending on the code that processes the dataset.  If I simply iterate over the dataset (`for sample in dataset:  ...`), I see painfully slow loading of the buffer and then a crash when memory is exhausted:
```
2019-10-02 17:35:16.297163: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 1 of 15000
2019-10-02 17:35:23.878243: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 33 of 15000
2019-10-02 17:35:31.980430: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 65 of 15000
2019-10-02 17:35:47.316917: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 129 of 15000
2019-10-02 17:35:55.363270: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 161 of 15000
2019-10-02 17:36:02.425192: I tensorflow/core/kernels/data/shuffle_dataset_op.cc:143] Filling up shuffle buffer (this may take a while): 193 of 15000
Killed
```
If I pass the dataset into model.fit_generator, it runs over 96 batches of input (3 files worth of data at 1024 samples per file == 96*32) and then stops with:
[...]

Issue #396, opened 2019-06-21
iterator from make_reader hangs after 10 epochs even if num_epochs=None


sdegryze:
In the example below, I'm iterating over the MNIST parquet file generated by [the MNIST example](https://github.com/uber/petastorm/blob/master/examples/mnist/generate_petastorm_mnist.py)

`num_epochs` is set to `None` to get the "infinite number of epochs" behavior.

However, the iterator loop hangs after 600 batches. Given that the parquet file contains 60000 records and I requested 1000 as my batch size, this corresponds to 10 epochs.

Code to reproduce:

```python
import tensorflow as tf
from petastorm import make_reader
from petastorm.tf_utils import make_petastorm_dataset


def streaming_parser(serialized_example):   
    image_data = tf.cast(tf.reshape(serialized_example.image, [784]), tf.float32)
    label = serialized_example.digit
    return {"image_data": image_data}, label


with make_reader("s3://path/to/train",
                 num_epochs=None,
                 cur_shard=0,
                 shard_count=1) as reader:
    exp_dataset = (make_petastorm_dataset(reader)
                   .map(streaming_parser)
                   .batch(1000))

features, labels = exp_dataset.make_one_shot_iterator().get_next()

with tf.Session() as sess:
    sess.run(tf.local_variables_initializer())
    sess.run(tf.global_variables_initializer())
    cum_count = 0
    for idx in range(610):
        labels_manifested = sess.run([labels])
        count = labels_manifested[0].shape[0]
        cum_count += count
        print(f"Batch {idx}, contains {count} records, total records read is {cum_count}")
```

selitvin:
Please try adjusting the scope of the `make_reader` context manager. I think that in your case, the reader is actually stopped because you exit the context manager, which triggers earlier then expected epoch termination.

This code works for me:
```python
import tensorflow as tf
from petastorm import make_reader
from petastorm.tf_utils import make_petastorm_dataset


def streaming_parser(serialized_example):
    image_data = tf.cast(tf.reshape(serialized_example.image, [784]), tf.float32)
    label = serialized_example.digit
    return {"image_data": image_data}, label


with make_reader("file:///tmp/mnist/train",
                 num_epochs=1000,
                 workers_count=1) as reader:
    exp_dataset = (make_petastorm_dataset(reader)
                   .map(streaming_parser)
                   .batch(1000))

    features, labels = exp_dataset.make_one_shot_iterator().get_next()

    with tf.Session() as sess:
[...]

PR #433, opened 2019-10-07
Raise an error if a user tries to read from a reader after it was stopped

selitvin:
Encountered this issue in #432 and #396. User tried reading after their context manager
has exited, which resulted in unexpected behavior.

### DIFF
--- a/petastorm/tests/test_end_to_end.py
+++ b/petastorm/tests/test_end_to_end.py
@@ -778,3 +778,20 @@ def test_multithreaded_reads(synthetic_dataset):
             results = [f.result() for f in futures]
             assert len(results) == len(synthetic_dataset.data)
             assert set(r.id for r in results) == set(d['id'] for d in synthetic_dataset.data)
+
+
+def test_should_fail_if_reading_out_of_context_manager(synthetic_dataset):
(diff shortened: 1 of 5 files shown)


petastorm/tests/test_end_to_end.py
```diff
@@ -768,3 +768,20 @@ def read_one_row():
...
             assert len(results) == len(synthetic_dataset.data)
             assert set(r.id for r in results) == set(d['id'] for d in synthetic_dataset.data)
+
+
+def test_should_fail_if_reading_out_of_context_manager(synthetic_dataset):
+    with make_reader(synthetic_dataset.url, workers_count=1) as reader:
+        next(reader)
+
+    with pytest.raises(RuntimeError, match='Trying to read a sample.*'):
+        next(reader)
```
Ivan-Dimitrov: Shouldn't this error be that the `reader` variable is undefined since `reader` went out of scope?

selitvin: Apparently, the variable is still defined after the context block. 
```
In [1]: with open('/tmp/zzz', 'wt') as f:
   ...:     pass
   ...: 

In [2]: f
```

PR #438, opened 2019-10-14
Support pyarrow 0.15 API

selitvin:


petastorm/compat.py
```diff
@@ -0,0 +1,61 @@
...
+        return column.data.num_chunks
+    else:
+        return column.num_chunks
+
+
+def compat_make_parquet_piece(path, open_file_func, **kwargs):
+    if _PYARROW_BEFORE_013:
+        return pq.ParquetDatasetPiece(path, **kwargs)
+    else:
+        return pq.ParquetDatasetPiece(path, open_file_func=open_file_func,   # pylint: disable=unexpected-keyword-arg
```
Ivan-Dimitrov: So for pyarrow > .13 all the reading helpers removed the open_file_func argument, except ParquetDatasetPiece added the argument ?

petastorm/compat.py
```diff
@@ -0,0 +1,61 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pyarrow as pa
+from packaging import version
+from pyarrow import parquet as pq
+
+_PYARROW_BEFORE_013 = version.parse(pa.__version__) < version.parse('0.13.0')
+
+
```
Ivan-Dimitrov: The reason for the file is fairly clear, but maybe a add a top line comment for documentation reasons.

petastorm/arrow_reader_worker.py
```diff
@@ -44,29 +45,36 @@ def read_next(self, workers_pool, schema, ngram):
...
+                else:
+                    column_as_numpy = column_as_pandas
+
                 if pa.types.is_string(column.type):
-                    result_dict[column.name] = column_as_pandas.astype(np.unicode_)
+                    result_dict[column_name] = column_as_numpy.astype(np.unicode_)
                 elif pa.types.is_list(column.type):
                     # Assuming all lists are of the same length, hence we can collate them into a matrix
-                    list_of_lists = column_as_pandas
+                    list_of_lists = column_as_numpy
```
Ivan-Dimitrov: Is the switch from pandas to numpy due to list issue in 0.15.0, or is there another reason

jsgoller1: :+1:  to this question

petastorm/compat.py
selitvin: Yep. It has moved. You configure ParquetDatasetPiece with the open-func once instead of passing a handle to each read.

xhochy:
Note that we have downstream integration tests in Arrow to detect incompatabilities much earlier, e.g. see https://github.com/apache/arrow/blob/master/dev/tasks/tasks.yml#L1553-L1568 It might be useful for you to also add a job for `petastorm` there.

PR #439, opened 2019-10-14
Adding np.string to str conversion in scalar codec's encode.

selitvin:


petastorm/tests/test_codec_scalar.py
```diff
@@ -76,3 +77,40 @@ def test_bad_unischema_field_shape():
...
+    assert isinstance(encoded, float)
+    assert expected == encoded
+
+def test_encode_scalar_string():
+    codec = ScalarCodec(StringType())
+    expected = 'surprise'
+    field = UnischemaField(name='field_string', numpy_dtype=np.unicode_, shape=(), codec=codec, nullable=False)
+    encoded = codec.encode(field, expected)
+    assert isinstance(encoded, str)
+    assert expected == encoded
```
Ivan-Dimitrov: I am confused by this equivalence? Is this a value comparison, object_id comparison? 

What is the point of encoding if there is no change? Is this equivalence only true for primitive types?

selitvin: PySpark understands only basic Python types when you try to write scalar fields into a parquet store (e.g. int, float). 

In order to have rich scalar type set (e.g. float32, float64, int8, etc...), we present numpy scalars to user. Codec will take these numpy types and convert them to Python types understandable by Spark.

Ivan-Dimitrov: okay understood

Issue #257, opened 2018-12-05
More unit test coverage for make_parquet_reader


selitvin:
Should extend `test_parquet_reader.py`. Some missing tests are
- subset of columns
- invalid column name;

PR #445, opened 2019-11-08
Added tests for test_parquet_reader.py

gregw18:
Added tests for selecting specific columns and requesting invalid columns to test_parquet_reader.
Modified specific column test to request specific column names, rather than regex patterns, so could select even columns rather than odd, so would always find at least one. Added comment explaining why regex patterns were a problem.

### DIFF
--- a/petastorm/tests/test_parquet_reader.py
+++ b/petastorm/tests/test_parquet_reader.py
@@ -145,7 +145,7 @@ def test_invalid_column_name(scalar_dataset, reader_factory):
         with pytest.raises(StopIteration):
             sample = next(reader)._asdict()
             # Have to do something with sample to avoid build error.
-            assert len(sample) == 0
+            assert sample == False
 
 
 @pytest.mark.parametrize('reader_factory', _D)
--- a/petastorm/tests/test_parquet_reader.py
+++ b/petastorm/tests/test_parquet_reader.py
@@ -145,7 +145,7 @@ def test_invalid_column_name(scalar_dataset, reader_factory):
         with pytest.raises(StopIteration):
             sample = next(reader)._asdict()
             # Have to do something with sample to avoid build error.
-            assert sample == False
+            assert not sample
 
 
 @pytest.mark.parametrize('reader_factory', _D)
--- a/petastorm/tests/test_parquet_reader.py
+++ b/petastorm/tests/test_parquet_reader.py
@@ -98,37 +98,6 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
 
 
 @pytest.mark.parametrize('reader_factory', _D)
-def test_invalid_column_name(scalar_dataset, reader_factory):
-    """Request a column that doesn't exist. Appears that when request only invalid fields,
-    DummyPool returns an EmptyResultError, which then causes a StopIteration in
-    ArrowReaderWorkerResultsQueueReader."""
-    all_fields = list(scalar_dataset.data[0].keys())
-    bad_field = _get_bad_field_name(all_fields)
-    requested_fields = [bad_field]
-
-    with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
-        with pytest.raises(StopIteration):
-            sample = next(reader)._asdict()
-            assert not sample
-
-
-@pytest.mark.parametrize('reader_factory', _D)
-def test_invalid_and_valid_column_names(scalar_dataset, reader_factory):
-    """Request one column that doesn't exist and one that does. Confirm that only get one field back and
-    that get exception when try to read from invalid field."""
-    all_fields = list(scalar_dataset.data[0].keys())
-    bad_field = _get_bad_field_name(all_fields)
-    requested_fields = [bad_field, all_fields[1]]
-
-    with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
-        sample = next(reader)._asdict()
-        assert len(sample) == 1
-        assert set(sample.keys()) == {all_fields[1]}
-        with pytest.raises(KeyError):
-            assert sample[bad_field] == ""
-
-
-@pytest.mark.parametrize('reader_factory', _D)
 def test_partitioned_field_is_not_queried(reader_factory, tmpdir):
     """Try datasets partitioned by a string, integer and string+integer fields"""
     url = 'file://' + tmpdir.strpath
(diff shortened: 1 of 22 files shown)


selitvin:
this is a good test. We can definitely add it.

I was considering to introduce #446. It may be a better behavior for the regex matching. What do you think?

gregw18:
That would definitely simplify my test, and better match how I expected that parameter to work, but my experience in this area is limited so my opinion isn't particularly informed! I hope that there aren't many people relying on the current behavior.

selitvin:
Yes. This is an unpleasant change short term and may break some people code. That being said, that's the right behavior to have long term.

gregw18:
It definitely follows the principle of least surprise, at least to a long time dos and windows programmer.
I'm also new to contributing to open source projects. Should I wait for #446 to be merged to master and then update my test?

selitvin:
Sorry for a delayed reply. Was travelling. I'll try to land 446 soon and have a release within couple of days.

selitvin:
Since #446  has landed on master, is ti ok to close this one?

gregw18:
I apologize for my confusing commit messages, but the purpose of these tests was to address #257 - I just happened to run into and get surprised by the schema_fields behavior. I hope that these tests are still useful.

petastorm/tests/test_parquet_reader.py
```diff
@@ -86,3 +85,35 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
         row_ids_batched = [row.id for row in reader]
     actual_row_ids = list(itertools.chain(*row_ids_batched))
     assert len(data) == len(actual_row_ids)
+
+
+@pytest.mark.parametrize('reader_factory', _D)
+def test_column_subset(scalar_dataset, reader_factory):
```
selitvin: Are we covering here functionality not already covered by `test_specify_columns_to_read` test in this file?

petastorm/tests/test_parquet_reader.py
```diff
@@ -86,3 +85,35 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
...
+def test_invalid_column_name(scalar_dataset, reader_factory):
+    """Request a column that doesn't exist, confirm that get exception."""
+    # Grab first field from expected dataset, append random characters to it to get an invalid field name.
+    all_fields = list(scalar_dataset.data[0].keys())
+    bad_field = all_fields[0]
+    while bad_field in all_fields:
+        bad_field += "VR46"
+    requested_fields = [bad_field]
+
+    with pytest.raises(StopIteration):
```
selitvin: `StopIteration` is not really an error, but just an indicator of the end of the data. Would `ValueError` be more appropriate here?

petastorm/tests/test_parquet_reader.py
gregw18: It was a duplicate test - I've removed it.

petastorm/tests/test_parquet_reader.py
gregw18: When requesting only invalid fields, it seems that DummyPool returns an EmptyResultError that ArrowReaderWorkerResultsQueueReader turns into a StopIteration exception. Is there a better way to test this, or should this type of request return something different?

selitvin:
Should we continue working on this PR? If yes, can you please rebase?

gregw18:
Updating this and #444 have been on my list, but I needed the reminder...

petastorm/tests/test_parquet_reader.py
```diff
@@ -90,6 +97,37 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
...
+def test_invalid_column_name(scalar_dataset, reader_factory):
+    """Request a column that doesn't exist. Appears that when request only invalid fields,
+    DummyPool returns an EmptyResultError, which then causes a StopIteration in
+    ArrowReaderWorkerResultsQueueReader."""
+    all_fields = list(scalar_dataset.data[0].keys())
+    bad_field = _get_bad_field_name(all_fields)
+    requested_fields = [bad_field]
+
+    with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
+        with pytest.raises(StopIteration):
```
selitvin: nit:
```python
	        with pytest.raises(StopIteration):
            next(reader)._asdict()
```
shouldn't trigger a build error

petastorm/tests/test_parquet_reader.py
```diff
@@ -90,6 +97,37 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
...
+    """Request one column that doesn't exist and one that does. Confirm that only get one field back and
+    that get exception when try to read from invalid field."""
+    all_fields = list(scalar_dataset.data[0].keys())
+    bad_field = _get_bad_field_name(all_fields)
+    requested_fields = [bad_field, all_fields[1]]
+
+    with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
+        sample = next(reader)._asdict()
+        assert len(sample) == 1
+        with pytest.raises(KeyError):
```
selitvin: A bit more explicit check that also covers correctness of querying all_fields[1] could be:
```python
    with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
        sample = next(reader)._asdict()
        assert set(sample.keys()) == {all_fields[1]}
```

petastorm/tests/test_parquet_reader.py
gregw18: I removed the stale comment.

petastorm/tests/test_parquet_reader.py
gregw18: Thanks - I've added that assertion. (I also kept the KeyError check, to show what the expected result of requesting an invalid field is.)

selitvin:
Can you please rebase? Have a conflict with master for some reason...

gregw18:
Please let me know if this still generates a conflict, or is just too ugly (5000 commits for 2 functions!) and I will cut and paste the changes into a new PR. (I've learned a lot about git since starting this and hopefully will be cleaner going forward.)

selitvin:
Hmmm. Github  still see a conflict. You don't see the same message?
image

gregw18:
Thanks for the screenshot - I had assumed that I wasn't looking in the right spot for the error.
Github is telling me that there are no conflicts, with the caveat that I don't have write access and so can't merge pull requests.
image
I hope that it's ok, but I've created a clean PR, #486, to replace this one. Hopefully it will let you land this and not spend any more time on it!

selitvin:
Awesome! Landed the new PR. Thanks for the tests!

PR #446, opened 2019-11-08
Regex specified in `make_reader`'s `schema_fields` argument now must match the entire field name.

selitvin:
Before this change, even a substring match was considered a match. For example:

Before this change:
- field name: 'field_123' would be loaded from the dataset with `make_reader(..., schema_fields=['123'])`.

After this change, user will need to specify a regex to match the full field name, for example:
- `schema_fields=['field_123']`
- `schema_fields=['.*123']`

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -401,19 +401,41 @@ def insert_explicit_nulls(unischema, row_dict):
                 raise ValueError('Field {} is not found in the row_dict, but is not nullable.'.format(field_name))
 
 
+def _fullmatch(regex, string, flags=0):
+    """Emulate python-3.4 re.fullmatch()."""
+    if six.PY2:
+        m = re.match(regex, string, flags=flags)
+        if m and (m.span() == (0, len(string))):
+            return m
+    else:
+        return re.fullmatch(regex, string, flags)
+
+
 def match_unischema_fields(schema, field_regex):
     """Returns a list of :class:`~petastorm.unischema.UnischemaField` objects that match a regular expression.
 
     :param schema: An instance of a :class:`~petastorm.unischema.Unischema` object.
-    :param field_regex: A list of regular expression patterns.
+    :param field_regex: A list of regular expression patterns. A field is matched if the regular expression matches
+      the entire field name.
     :return: A list of :class:`~petastorm.unischema.UnischemaField` instances matching at least one of the regular
       expression patterns given by ``field_regex``.
     """
     if field_regex:
         unischema_fields = []
+        legacy_unischema_fields = []
         for pattern in field_regex:
             unischema_fields.extend(
-                [field for field_name, field in schema.fields.items() if re.match(pattern, field_name)])
+                [field for field_name, field in schema.fields.items() if _fullmatch(pattern, field_name)])
+            legacy_unischema_fields.extend([field for field_name, field in schema.fields.items()
+                                            if re.match(pattern, field_name)])
+        if unischema_fields != legacy_unischema_fields:
+            field_names = {f.name for f in unischema_fields}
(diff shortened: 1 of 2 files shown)


petastorm/unischema.py
```diff
@@ -401,19 +401,27 @@ def insert_explicit_nulls(unischema, row_dict):
                 raise ValueError('Field {} is not found in the row_dict, but is not nullable.'.format(field_name))
 
 
+def _fullmatch(regex, string, flags=0):
```
jsgoller1: Since Petastorm is supported for both Python 2 and 3, won't this wind up using doing the emulated version even when `re.fullmatch()` is supported? My worry here is that there might be weird edgecases that `fullmatch()` handles (regular expressions are complicated, so I think this is a reasonable concern) and our implementation might not. Could we change the implementation here to something like:
```
def _fullmatch(regex, string, flags=0):
  if `'fullmatch' in dir(re):
    re.fullmatch(regex, string, flags)
  else:
    # existing code below 
```

petastorm/unischema.py
```diff
@@ -401,19 +401,27 @@ def insert_explicit_nulls(unischema, row_dict):
...
+    """Emulate python-3.4 re.fullmatch()."""
+    m = re.match(regex, string, flags=flags)
+    if m and m.span() == (0, len(string)):
```
jsgoller1: Could we use explicit parentheses here, i.e. `if m and (m.span() == (0, len(string))):`

Ivan-Dimitrov:
I am a fan of using proper regex matching. This could end up being a breaking change for users though.

selitvin:
> I am a fan of using proper regex matching. This could end up being a breaking change for users though.

Yep. This could be a breaking change. Current behavior though is too confusing, IMO. Imaging having `field1` and `field10`. Then `['field1']` would actually match both. I'll add a warning if the old and the new method does not match. That should mitigate partially the fact that this is a breaking change.

petastorm/unischema.py
```diff
@@ -401,19 +401,41 @@ def insert_explicit_nulls(unischema, row_dict):
                 raise ValueError('Field {} is not found in the row_dict, but is not nullable.'.format(field_name))
 
 
+def _fullmatch(regex, string, flags=0):
```
Ivan-Dimitrov: Missing docstring

Ivan-Dimitrov:
tested the behavior, seems solid.

Thanks for adding the warning.

petastorm/unischema.py
```diff
@@ -401,19 +401,41 @@ def insert_explicit_nulls(unischema, row_dict):
...
     :return: A list of :class:`~petastorm.unischema.UnischemaField` instances matching at least one of the regular
       expression patterns given by ``field_regex``.
     """
     if field_regex:
         unischema_fields = []
+        legacy_unischema_fields = []
         for pattern in field_regex:
             unischema_fields.extend(
-                [field for field_name, field in schema.fields.items() if re.match(pattern, field_name)])
+                [field for field_name, field in schema.fields.items() if _fullmatch(pattern, field_name)])
```
jsgoller1: Nit: why are we using `.extend()` here? Could we just say `legacy_unischema_fields =  [field for field_name, field in schema.fields.items() if _fullmatch(pattern, field_name)]`?

petastorm/unischema.py
selitvin: Done

petastorm/unischema.py
selitvin: Done

petastorm/unischema.py
selitvin: We are going over a list of regular expression. For each one, we need extend the `unischema_fields` list. In your code samples, we would end up constantly overwriting it and use only the last one.

PR #450, opened 2019-11-25
Use ubuntu 18.04 as CI image

selitvin:
WIP

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -5,36 +5,32 @@ all: rm-image build
 # pip. If the install process is killed midway through,
 # rerun it and check `docker stats` during execution; you may
 # need to increase Docker's memory limits.
+TAG := $(shell date +%Y-%m-%d_%H-%M-%S)
+IMAGE := "docker.io/selitvin/petastorm_ci"
 build:
 	cd .. && docker build . \
 	-f ./docker/Dockerfile \
-	--tag petastorm
-
-# Create the CI container for Travis use
-push_ci_container:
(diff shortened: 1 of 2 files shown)


docker/Makefile
```diff
@@ -5,36 +5,32 @@ all: rm-image build
...
+	@echo
+	@echo "    docker push $(IMAGE):$(TAG)"
 
 # Run a container to serve as a development workspace with
 # the local repo mounted into it. IMPORTANT: this gives
 # the container access to the host's filesystem!
 run:
 	docker run \
-	-dt \
+	-it \
```
jsgoller1: Do we not want to run in daemon mode anymore (`docker run -d`)?

petastorm/tests/test_disk_cache.py
```diff
@@ -54,7 +54,7 @@ def test_size_limit_constraint(tmpdir):
         cache.get('some_key_{}'.format(i), lambda: a_record)
 
     # Check that we are more or less within the size limit
-    assert _recursive_folder_size(tmpdir.strpath) < 2 * MB
+    assert _recursive_folder_size(tmpdir.strpath) < 3 * MB
```
jsgoller1: This change is also in #452, FYI. Not sure if these branches should be stacked or not.

docker/Makefile
selitvin: For debugging code in the container, I found it useful to just `make run` and not two steps, starting a daemon and then execing a shell. What do you think?

PR #452, opened 2019-11-28
Upgrade petastorm CI image to 2019-11-27_13-25-17. Uses ubuntu 18.04.

selitvin:
Also made some fixups to support torch 1.3 properly.


docker/Makefile
```diff
@@ -24,7 +24,7 @@ run:
 	-dt \
 	-v `pwd`/..:/petastorm \
 	--name petastorm \
-	petastorm
+	petastorm_ci:latest
```
jsgoller1: Should this be `selitvin/petastorm_ci:latest`?

selitvin: Removed from this PR

PR #459, opened 2019-12-12
update materialize_dataset documentation for filesystem factory arg

jakelarkn:
as title


jakelarkn:
Thanks.  I'm still having some issues with this API, when you have a chance could look at my issue in the petastorm channel (uchat)?

Issue #367, opened 2019-05-03
No need to specify `removed` argument in `TransformSpec`


selitvin:
We can deduce the list of the removed fields by examining return value of the user transform function:

```python
def user_transform(row):
   del row['deleted_field']
```

gregw18:
I've been looking at some Unischema checking issues - #297, #355, #367 and #402 - and have started work on supporting adding and removing fields. I would like to confirm that I'm heading in the right direction. My current goals are:

1. When running under TensorFlow, emit helpful warnings before crashing - since can't change outputs after graph is created.

2. When not running under TensorFlow, want to work with whatever fields we're given.

3. This is just my opinion, but I'm not sure how much value there is in allowing changing the type of existing fields in a Unischema. Adding a new field (with a new name) and removing the old one would seem to give equivalent functionality, and not require any additional code. (Changing the contents of an existing field, without changing its type, doesn't require any changes to the schema and so should continue to be fine.)

Does this approach sound acceptable?

selitvin:
Thank you for helping with this!

Not sure I fully understand the context. The schema exposed by the Reader object already includes adjustments declared by a `TransformSpec`. In this context I am not sure I understand:

> 1. When running under TensorFlow, emit helpful warnings before crashing - since can't change outputs after graph is created.
Is this achieved by validating the sample at the runtime against the schema?

> started work on supporting adding and removing fields
To clarify: which classes are affected? When fields are added/removed? Does this mean that the schema could change per-returned-sample?

Can you please elaborate, which classes are expected to change (and how)?

gregw18:
Sorry - I've been staring at these issues for too long and lost their context.
I believe that all four of the issues I mentioned are discussing when the user defines a TransformSpec.func that makes changes (adds or removes fields) that they didn't list in edit_fields or removed_fields. The general idea, I inferred, was that if running under TensorFlow, we can't handle this - we will crash, but we could be a bit friendlier and emit some helpful excuses before we die. If not running under Tensorflow, as long as we update the appropriate Unichema instance, based on the data returned from the TransformSpec.func, it should be possible to work, and the user doesn't really even need to populate the edit_fields and removed_fields elements of TransformSpec.
At the moment I've just hacked together a proof of concept (including tests) to prove to myself that I can do this - it's not handling all cases yet. If we agree on the goals, I would start with a clean branch, move over the required code (you wouldn't believe how many print statements I've scattered around to help understand how this is working) and then finish implementing.
[...]

Issue #402, opened 2019-06-27
Values returned by a transform function are not validated against the schema


selitvin:

praateekmahajan:
Also, is there a way to use transform function to change the schema using the `transform_spec`?

```
def func(row)
  row['existing_col'] = row['existing_col'].apply(tokenize_me)
  row['new_col'] = row['existing_col'].apply(lambda x : some_func(x))
  
  return row
transform = TransformSpec(func)
```

Is there a way we add `new_col` on the fly to the data?

selitvin:
You are using this function from PyTorch? (For TF, the type needs to be explicit as we are creating the graph before we read any sample). We planned to do #355, is this the same ask?

praateekmahajan:
I am not sure if the ask is the same. It might be related, but I don't see when does `TransformSpec` results in a failure for PyTorch. What I am curious is how to use `TransformSpec` for a `batch_reader` to create new columns on the fly. 

I think reading through the following example should be trivial
```
def _transform_row(df):
    import pandas as pd
    df['new_col'] = pd.Series(100)
    return df

transform = TransformSpec(func=_transform_row)

train_loader = DataLoader(reader = make_batch_reader(petastorm_dataset_url, 
transform_spec=transform))
for i, data in enumerate(train_loader):
  print(data.keys())
  assert 'new_col' in data.keys(), "New col wasn't created"
  break 
```

And this code would result in `AssertionError : New col wasn't created`


For the curious ones, a use case could be, you have a text column and you'd want to generate multiple features (for eg BERT tokens, masks, padding based on text column)

selitvin:
I think what's missing is `edit_fields=...` argument to the `TensorSpec`. Originally this explicit specification for the new field is a TF requirement. I think it should be relaxed for pytorch to avoid the confusion (assuming I understood your question correctly and there is a confusion).

Here is a modified `examples/hello_world/external_dataset/python_hello_world.py` I used to demonstrate the feature. Hope I understood your question correctly, but please let me know if I missed something:

```python
from __future__ import print_function

from petastorm import make_batch_reader, TransformSpec
from petastorm.pytorch import DataLoader
from petastorm.unischema import UnischemaField
import numpy as np

def _transform_row(df):
    import pandas as pd
    df['new_col'] = pd.Series((10,20,30,40,50))
    return df

transform = TransformSpec(func=_transform_row, edit_fields=[UnischemaField('new_col', np.float64, (), None, False)])

[...]

Issue #355, opened 2019-04-10
Make it simpler for PyTorch users to use TransformSpec


selitvin:
If a new field is added/modified in the transform function and the change is not documented in the `edit_fields`/`removed_fields` arguments of the `TransformSpec`, the code would in Unischema's make_named_tuple. For a TF case we have to know what are the exact tensor types, so we can wire them into the TF graph. However, for PyTorch users, there is no real value. 

Can we make the code not crash for the PyTorch users in this case?

Issue #297, opened 2019-01-23
Add validation of the shapes/types returned by transform function specified by user


selitvin:
When user specifies a transform, they specify a transform function and an expected output scheme. If the data returned by a user has a mismatching shape/type, the code will fail somewhere downstream with an unclear error message in case of TF. We should add a check that verifies compliance of the data returned by the user with the schema specified by the user in the `TransformSpec` object.

PR #465, opened 2020-01-04
Make data transform logic a bit easier for non-TensorFlow users.

gregw18:
Previously, when using make_reader*'s transform_spec argument, both TensorFlow and non-TensorFlow users had to specify what changes they were going to make to the fields returned by the reader, even though only TensorFlow needed to know ahead of time. Now, non-TensorFlow users can just provide a transform function. Hopefully resolves issues #367, #297, #355 and #402.

### DIFF
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -24,8 +24,35 @@ from pyarrow.parquet import ParquetFile
 
 from petastorm.cache import NullCache
 from petastorm.compat import compat_piece_read, compat_table_columns_gen, compat_column_data
+from petastorm.transform import TransformSpec
+from petastorm.unischema import make_namedtuple
 from petastorm.workers_pool import EmptyResultError
 from petastorm.workers_pool.worker_base import WorkerBase
+
+
+def _apply_transform(data_as_pandas, transform):
+    """Applies transform (either TransformSpec.func (if specified) or transform if it is a callable) to provided
+    data. Removes all fields as specified by TransformSpec.removed_fields, unless the fields were already
+    deleted by transform.
+    """
+    if isinstance(transform, TransformSpec):
+        # A user may omit `func` value if they intend just to delete some fields using the TransformSpec
+        if transform.func:
+            transformed_result = transform.func(data_as_pandas)
+        else:
+            transformed_result = data_as_pandas
+
+        # If transform function left a field that is listed in transform_spec's remove_fields, we remove it
+        # ourselves. Allows for the following transform-spec objects to be created:
+        # TransformSpec(removed_fields=['some_field'])
+        for field_to_remove in set(transformed_result.columns) & set(transform.removed_fields):
+            del transformed_result[field_to_remove]
+    elif callable(transform):
+        transformed_result = transform(data_as_pandas)
+    else:
+        transformed_result = data_as_pandas
+
+    return transformed_result
 
 
 class ArrowReaderWorkerResultsQueueReader(object):
@@ -36,7 +63,7 @@ class ArrowReaderWorkerResultsQueueReader(object):
     def batched_output(self):
         return True
 
-    def read_next(self, workers_pool, schema, ngram):
+    def read_next(self, workers_pool, ngram):
         try:
             assert not ngram, 'ArrowReader does not support ngrams for now'
 
@@ -76,7 +103,7 @@ class ArrowReaderWorkerResultsQueueReader(object):
                 else:
                     result_dict[column_name] = column_as_numpy
 
-            return schema.make_namedtuple(**result_dict)
+            return make_namedtuple(**result_dict)
 
         except EmptyResultError:
             raise StopIteration
@@ -92,7 +119,7 @@ class ArrowReaderWorker(WorkerBase):
         self._ngram = args[3]
         self._split_pieces = args[4]
         self._local_cache = args[5]
-        self._transform_spec = args[6]
+        self._transform = args[6]
 
         if self._ngram:
             raise NotImplementedError('ngrams are not supported by ArrowReaderWorker')
@@ -164,20 +191,9 @@ class ArrowReaderWorker(WorkerBase):
 
         result = self._read_with_shuffle_row_drop(piece, pq_file, column_names, shuffle_row_drop_range)
 
-        if self._transform_spec:
+        if self._transform:
             result_as_pandas = result.to_pandas()
-            # A user may omit `func` value if they intend just to delete some fields using the TransformSpec
-            if self._transform_spec.func:
-                transformed_result = self._transform_spec.func(result_as_pandas)
-            else:
-                transformed_result = result_as_pandas
-
-            # If transform function left a field that is listed in transform_spec's remove_fields, we remove it
-            # ourselves. Allows for the following transform-spec objects to be created:
-            # TransformSpec(removed_fields=['some field'])
-            for field_to_remove in set(transformed_result.columns) & set(self._transform_spec.removed_fields):
-                del transformed_result[field_to_remove]
-
+            transformed_result = _apply_transform(result_as_pandas, self._transform)
             result = pa.Table.from_pandas(transformed_result, preserve_index=False)
 
         return result
@@ -238,8 +254,8 @@ class ArrowReaderWorker(WorkerBase):
 
         result = result_data_frame[match_predicate_mask]
 
-        if self._transform_spec:
-            result = self._transform_spec.func(result)
+        if self._transform:
+            result = _apply_transform(result, self._transform)
 
         return pa.Table.from_pandas(result, preserve_index=False)
 
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -34,7 +34,7 @@ from petastorm.reader_impl.arrow_table_serializer import ArrowTableSerializer
 from petastorm.reader_impl.pickle_serializer import PickleSerializer
 from petastorm.reader_impl.pyarrow_serializer import PyArrowSerializer
 from petastorm.selectors import RowGroupSelectorBase
-from petastorm.transform import transform_schema
+from petastorm.transform import transform_schema, TransformSpec
 from petastorm.workers_pool.dummy_pool import DummyPool
 from petastorm.workers_pool.process_pool import ProcessPool
 from petastorm.workers_pool.thread_pool import ThreadPool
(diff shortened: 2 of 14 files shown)


petastorm/unischema.py
```diff
@@ -269,24 +287,69 @@ def as_spark_schema(self):
...
+                          + str(unexpected_fields)
+                if msg:
+                    msghdr = "It seems that the transform function changed the fields that are returned, and these " \
+                             "changes weren't listed in the TransformSpec.edit_fields and removed_fields properties." \
+                             "If running under TensorFlow this will result in a crash, so please use these " \
+                             "parameters to avoid this crash. If not running under TensorFlow, please feel free to " \
+                             "ignore this warning.\n"
+                    warnings.warn(msghdr + msg)
+
+                self._is_verified = True
```
selitvin: Unclear why the state of verification of a particular `result` instance is stored in the schema? It could be that:
```python
some_schema.verify_fields(good_data) == True
# and then....
some_schema.verify_fields(bad_data) == True  # while should be False
```

petastorm/unischema.py
```diff
@@ -269,24 +287,69 @@ def as_spark_schema(self):
...
+        """
+        if "_is_verified" not in self.__dict__:
+            # Can happen if read older version from a file.
+            self._is_verified = False
+
+        if not self._is_verified:
+            # Can only verify if have some data.
+            if results:
+                actual_field_names = self._get_field_names(results)
+                missing_fields = []
```
selitvin: Working with sets could be clearer and possibly faster, as we would avoid python loops and n^2 complexity. Something like this?
```python
actual = set(self._get_field_names(results))
expected  = set(self._fields.keys())
missing = expected - actual
unexpected= actual - expected
```

petastorm/unischema.py
```diff
@@ -269,24 +287,69 @@ def as_spark_schema(self):
...
+                # If received a dictionary, rather than a list of keys, grab keys from each item in the dictionary,
+                # then remove any duplicates.
+                field_names = []
+                for key in results.keys():
+                    field_names = field_names + list(results[key]._fields)
+                # Remove duplicates.
+                field_names = list(set(field_names))
+        return field_names
+
+    def verify_fields(self, results, expected_fields=None):
```
selitvin: IMO, `results` is somewhat confusing here. It may be `results` when used in some context, e.g. what we doing here, but in other cases it could not "results" but something else.

petastorm/reader.py
```diff
@@ -607,7 +646,10 @@ def __next__(self):
                                'make_reader/make_batch_reader context manager has exited but you try to '
                                'fetch a sample from it anyway')
         try:
-            return self._results_queue_reader.read_next(self._workers_pool, self.schema, self.ngram)
+            results = self._results_queue_reader.read_next(self._workers_pool, self.ngram)
+            if '_schema' in self.__dict__:
```
selitvin: It might be better to (a) check if the Transform_spec.func returned proper types right after we run it on a worker thread/process and (b) only in case that we know that we work with Transform_spec and not a callable while we raise errors and not warnings. My reasoning is:
(a): Main thread/process is an expensive place to run the code - running it on the worker leaves more opportunity for parallelization.
(b): Messing up types with TransformSpec is a big deal - user promissed us something and violated it, hence "deserves" to crash. In the "transform is callable"  case, user has a carte blache to do whatever they want with the types, so warning seems inappropriate.

gregw18:
Thanks for the clear and thoughtful review! I apologize for making you feel that you had to spend your Sunday night doing this ...
With one exception, I think that I understand your suggestions and agree with them. I had an unstated assumption when I was working on this - that I only needed to check the first batch of data returned using a given Unischema, because all data passing through a given Reader would be the same format. My reasoning, based on my limited experience with machine learning systems, was that you have to feed them the same format of data in each batch. Thus, the data coming from disk would always be the same format, and, while it would certainly be possible to return different fields from a transform function based on arbitrary conditions, it would be self-defeating to do so, and thus checking the format for each batch would be unnecessary overhead.
Can you can confirm that it would be best to check every set of data returned from a given reader?
If you're happy with the transform parameter change to the `make_reader` calls, I'll also update the code examples to use transform rather than transform_spec, so that the documentation should be in sync.

selitvin:
Hey - definitely no need to apologize - it's me that should apologize for not reviewing this earlier. :)
I agree with your assumptions that the data format is not expected to change between consequent examples. My reasoning for actually making the check (I assume we are talking only about TransformSepc case, as `callable` we don't have a schema declaration anyway):
 # I did not check this edge case, so I might be wrong about it: with a parquet store with multiple files, the schema of the each file might be different: e.g. by-date partitions, and then the schema mutates at some date. The first batch might have correct type, but others (read from different parquet files) might end-up having different fields.
 # Would think that checking for a set of field names/types per batch (or per record, assuming rows are big, as in `make_reader` case) should be pretty cheap and go unnoticed.
 # Mutating schema, as in (1) can be handled by pytorch models and maybe desirable (as python code in the data reading layer may adapt to the schema changes.). 


I think your treatment of `transform_spec` / `transform` argument is good. Although we'll need to deprecate `transform_spec` in the future, it's better to do what you propose since it will lead us to a cleaner API.

petastorm/arrow_reader_worker.py
```diff
@@ -24,10 +24,37 @@
...
+
+        # If transform function left a field that is listed in transform_spec's remove_fields, we remove it
+        # ourselves. Allows for the following transform-spec objects to be created:
+        # TransformSpec(removed_fields=['some_field'])
+        for field_to_remove in set(transformed_result.columns) & set(transform.removed_fields):
+            del transformed_result[field_to_remove]
+    elif callable(transform):
+        transformed_result = transform(data_as_pandas)
+    else:
+        transformed_result = data_as_pandas
```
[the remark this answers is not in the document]
gregw18: Done.

gregw18:
Stepping back a bit, I would like to verify that what I’m trying to accomplish matches what you would like to see done. Again, I don’t have a personal direction that I want to push this project in – I just want to make sure that my limited experience in this area isn’t causing me to head in the wrong direction.
My original goals were: 1. To make doing transformations a bit easier for non-TF users, by not requiring them to create a full `TransformSpec`. 2. To make schema discrepancy errors for TF users more obvious, by giving a warning about the schema discrepancy before crashing.
From a usage perspective, non-TF users can always just use a transform function, regardless of whether they are adding or removing fields, and we should just work. For TF users, if they are not adding or removing fields they can also use a transform function, but if they are making schema changes, they need to use a full `TransformSpec` so that TF knows what fields will be returned.
In terms of checking returned data against the expected schema, if the user provided a full `TransformSpec`, I think we should check it regardless of whether we’re running under TF, as a discrepancy may be an inadvertent error. If they provided just a callable, if non-TF probably shouldn’t check – the user is telling us to trust them. If running under TF, and the callable does add or remove fields, I would still like to tell the user why we’re about to crash. (Even though the user should know better than to do this.)
If a given `reader` is returning different fields over time, for non-TF we should just work. However, for TF, even a `TransformSpec` can only specify one set of differences from the stored schema, which, with TF’s fixed graph, means that we’re going to crash. However, it would still be good to explain why beforehand.
All of which seems to indicate that I should check all data returned by a `reader`, not just the first batch, but I also need to be able to tell if running under TF or not – which I haven’t been able to figure out how to do.
Finally, if we do agree on checking a `TransformSpec` for the non-TF case, I would lean towards just checking the first batch of data. If there is a discrepancy, and it is intentional, or it is because the fields returned vary over time, I wouldn’t want to emit a huge number of identical warnings.
Thoughts?

selitvin:
Thank you for the detailed brain-dump!

I agree with all your points. My interpretation of what I understood:
1. TransformSpec provided --> always check compliance with output schema. if a user declared post-transform schema, he should comply (otherwise, it's likely have been an inadvertent error).
2.  callable: non-TF - no schema was declared, user can do whatever they want. Not sure we need to do an extra effort checking against the schema used for graph building, although we could (in `tf_utils.py` module, perhaps we could capture the schema and check as part of `py_func` body)

>  I would lean towards just checking the first batch of data...., I wouldn’t want to emit a huge number of identical warnings.
If we raise an error in (1), that wouldn't be the problem, right? Once user-transform did something non-compliant we abort the execution.

gregw18:
I've implemented most of your recommendations, but am struggling with hiding reader.schema when not using a TransformSpec and non-TF. I've considered two implementations:
1. Always saving the schema but hiding it in the non-TransformSpec case until tf_tensors is called, which would set a flag in the reader to make it visible. This assumes that tf_tensors is always called when using TensorFlow.
2. Passing an "is_tensorflow" parameter to make_reader and make_batch_reader.
However, I'm not clear what the benefit of hiding the schema is, and thus whether it outweighs the extra complexity of sometimes hiding it.

gregw18:
I'm struggling with implementing my last goal for this project - giving TF users an understandable warning/error if the data being returned doesn't match what we expect. Some options are given below, along with my opinion of each. Any feedback would be appreciated.
Also, reader.schema is still always available because it is used by tf_utils.py in various places. Was the goal of hiding it in some cases to avoid providing invalid data in cases where it doesn't match the returned data?
Options:
1. Submit  what I currently have – return an error if user provides a TransformSpec but the returned data doesn’t match what is expected. Otherwise, no longer crashes for non-TF if format doesn’t match expected, but TF still crashes, with a rather cryptic error.
2. Add a warning if using TF but not using a TransformSpec, by doing checking on main thread, using a flag set in dequeue_sample_impl.
3. Add a warning if using TF but not using a TransformSpec, using a flag set in dequeue_sample_impl, by creating separate thread to do checking, or by letting existing threads know that I want checking.
4. Add a new parameter to make_reader/make_batch reader, say check_returned_format, which reader.init passes to threads/processes (which do the checking), when created. If the parameter is not provided, default to true if TransformSpec or transform provided, false otherwise. Suggest in documentation that always be set to true if using TF.

Thoughts:
1. Easy to do, but doesn’t meet all of my goals – if not using TransformSpec, doesn’t explain data mismatch crashes to TF users.
2. Relatively easy to implement, albeit a bit ugly (have to do in `tf_utils.py` dequeue_sample_impl in _tf_tensors_nonngram and _tf_tensors_ngram, and validation logic has to handle the different data structures in the main thread vs in the worker_pool), and there are possible performance problems.
3. Not sure how or where to create new thread, how it should ensure that data is checked before it is passed on to reader.next. Definitely don’t want to create yet another stage to processing. Probably too late to modify parameters for existing threads/processes, as they will have already passed some data into the next stage.
4. Easy to implement. Consistent and obvious for callers, but is a change from current behavior.

PR #472, opened 2020-01-22
Support GCSFS

### LINKED ISSUE #419: Support for GCS
Would be great to also have support to read data from GCS paths, any plans on updating this. I see ludwig actually already has support for this!

megaserg:
Resolves #419

### DIFF
--- a/petastorm/fs_utils.py
+++ b/petastorm/fs_utils.py
@@ -16,6 +16,7 @@ import pyarrow
 import six
 from six.moves.urllib.parse import urlparse
 
+from petastorm.gcsfs_helpers.gcsfs_wrapper import GCSFSWrapper
 from petastorm.hdfs.namenode import HdfsNamenodeResolver, HdfsConnector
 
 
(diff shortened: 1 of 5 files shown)


megaserg:
> This is awesome - thanks for the contribution! Would you be able to provide some basic test coverage for the `GCSFSWrapper`?

I could probably port the test from `pyarrow` where [it tests S3FSWrapper](https://github.com/apache/arrow/blob/d98fc458f3941740782a470b1f4436bc8603f0f6/python/pyarrow/tests/test_parquet.py#L1885), but it would pull a few helper methods with it. Is this what you would like?

selitvin:
Seems like they actually have a full blown integration test against s3? It is nice, but I think that would be a bigger project to setup something like this for Petastorm. I was thinking more about mocking `gcsfs` classes. That being said, I am not sure there is much value in testing it at the unit level... so up to you. 

Let me know what your intent and I'll land it on master and bake a release right after.

megaserg:
Let's merge as is. If people report issues, then let's create a test against those :relaxed:

PR #474, opened 2020-01-27
Sending FINISHED message to workers when main process dies

ingolfured:
## The problem
If you are using Petastorm in process mode and the main process dies unexpectedly, it leaves the workers running until the user manually kills them. In some environments, this can be quite tricky, especially if you don't SSH access to the box.

## The solution
I propose a solution where we use a separate process to monitor the main process. If the main process dies unexpectedly, this new process, `ProcessMonitor` sends a `FINISHED` message to the alive workers via a new channel.

### DIFF
--- a/petastorm/workers_pool/constants.py
+++ b/petastorm/workers_pool/constants.py
@@ -1,0 +1,4 @@
+# When CONTROL_FINISHED is passed via control socket to a worker, the worker will terminate
+CONTROL_FINISHED = "FINISHED"
+SOCKET_LINGER_MS = 1000
+LOCALHOST = 'tcp://127.0.0.1'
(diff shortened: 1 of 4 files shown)


selitvin:
Do you think we have to have another process monitoring the master? Having an additional process in the system might introduce additional uncertainty and potentially flakiness (same kind of multiprocess system issue you are trying to address with this PR).

I am curious, did you consider to make workers terminate themselves. There could be several options, I guess:
* register to [get a signal](https://stackoverflow.com/questions/284325/how-to-make-child-process-die-after-parent-exits) when its parent dies and terminate itself
* Let workers monitor the parent process (e.g. running a monitor thread).

ingolfured:
Interesting, getting a signal when the parent dies works but only on Linux. Would that be ok?
Otherwise, option 2 seems more robust than the solution I proposed

ingolfured:
Closed since a better implementation is here: #482

PR #477, opened 2020-01-28
Fix incorrect counting of number of row-groups per piece.

### LINKED ISSUE #447: Error reading parquet files made by AWS Athena
I made a bunch of parquet files using an amazon athena CTAS query. I downloaded these files to first test locally (the end goal is to access the data from S3).

If I run the code below;

```python
import s3fs
from petastorm.reader import make_batch_reader
from petastorm.tf_utils import make_petastorm_dataset

dataset_url = "file:///Data/test-parquet"

with make_batch_reader(dataset_url) as reader:
    dataset = make_petastorm_dataset(reader)
    for batch in dataset:
        break
batch.correct
```

I receive a lot of warnings and then an error in `for batch in dataset`

`pyarrow.lib.ArrowIOError: The file only has 1 row groups, requested metadata for row group: 1`

If 1 look at dataset.take(1) or something alike, I do see the correct schema of the table. However, I don't seem to be able to access the data.

selitvin:
This bug manifests in either:
- `pyarrow.lib.ArrowIOError: The file only has <X> row groups, requested metadata for row group: <Y>` (where <X> and <Y> are integers) exception, or:
- silently - some rows in a dataset might have been silently ignored.

This issue occured when:
- using `make_batch_reader`
- number of row-groups in different parquet files (pieces) is different.

The issue was in introduced in petastorm 0.7.7

Resolves #447

### DIFF
--- a/petastorm/tests/test_common.py
+++ b/petastorm/tests/test_common.py
@@ -184,9 +184,12 @@ def create_test_scalar_dataset(output_url, num_rows, num_files=4, spark=None, pa
 
         spark = spark_session.getOrCreate()
         shutdown = True
+        hadoop_config = spark.sparkContext._jsc.hadoopConfiguration()
+        hadoop_config.setInt('parquet.block.size', 100)
 
     def expected_row(i):
         result = {'id': np.int32(i),
+                  'id_div_700': np.int32(i // 700),
                   'datetime': np.datetime64('2019-01-02'),
                   'timestamp': np.datetime64('2005-02-25T03:30'),
                   'string': np.unicode_('hello_{}'.format(i)),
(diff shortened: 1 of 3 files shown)


petastorm/tests/test_common.py
```diff
@@ -184,9 +184,12 @@ def create_test_scalar_dataset(output_url, num_rows, num_files=4, spark=None, pa
 
         spark = spark_session.getOrCreate()
         shutdown = True
+        hadoop_config = spark.sparkContext._jsc.hadoopConfiguration()
+        hadoop_config.setInt('parquet.block.size', 100)
 
     def expected_row(i):
         result = {'id': np.int32(i),
+                  'id_div_700': np.int32(i // 700),
```
jsgoller1: What is the significance of `700` here?

WeichenXu123:
Good work!

PR #485, opened 2020-02-03
Fixing sources of various deprecation warnings

selitvin:


jsgoller1:
@selitvin are you still intending to land this? If so, could you update the summary so I can review?

PR #490, opened 2020-02-16
assign transform result in _load_rows_with_predicate

jgblight:
I noticed that TransformSpecs were not being applied if there was also a row predicate specified, this PR updates the `_load_rows_with_predicate` function to match https://github.com/uber/petastorm/blob/master/petastorm/py_dict_reader_worker.py#L184


selitvin:
Thank you for the contribution - it looks good ! Please let me know if you would be able modify unit tests to make sure the code does not regress in the future. If not, I can help with that.

jgblight:
@selitvin Does this test look ok? Thanks for reviewing!

PR #492, opened 2020-02-24
New PyTorch DataLoader implementation using batched operations

fps7806:
The current DataLoader implementation relies on python iteration over individual rows of a parquet dataset. This is reasonable for small datasets but runs into throughput limitations:

Previous timing:
```
Samples per second for batch 10: 7.556e+04
Samples per second for batch 100: 1.001e+05
Samples per second for batch 1000: 1.083e+05
Samples per second for batch 100000: 4.364e+04
```

The new approach converts data into torch format BEFORE appending to the shuffling buffer and then relies on batched operations to perform all data appending/sampling.

New timing:
```
Samples per second for batch 10: 4.363e+05
Samples per second for batch 100: 1.468e+06
Samples per second for batch 1000: 1.607e+06
Samples per second for batch 100000: 1.141e+06
```
New timing with GPU-buffering:
```
Samples per second for batch 10: 3.87e+05
Samples per second for batch 100: 1.95e+06
Samples per second for batch 1000: 2.307e+06
Samples per second for batch 100000: 5.053e+06
```

In the best-case scenario, the samples/s increases 2 orders of magnitude. Although image-based problems would likely not achieve this throughput, these numbers are more representative of the throughput on tabular-data problems.

The proposed changes can be used out-of-the-box in most scenarios.

There are still many improvements that can be made to the ShufflingBuffer logic, but this implementation provides an initial step in that direction.

### DIFF
--- a/examples/timing.py
+++ b/examples/timing.py
@@ -1,0 +1,48 @@
+import time
+import numpy as np
+from petastorm.pytorch import DataLoader
+from collections import namedtuple
+import torch
+from functools import partial
+import sys
+
+
+class DummyReader(object):
+    @property
+    def is_batched_reader(self):
+        return True
+
+    def stop(self):
+        pass
+
+    def join(self):
+        pass
+
+    def __iter__(self):
+        nt = namedtuple("row", ["test"])
+        batch = nt(np.random.rand(1000, 64).astype(np.float32))
+        while True:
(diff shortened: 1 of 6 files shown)


fps7806:
Closed in favor of https://github.com/uber/petastorm/pull/540

PR #495, opened 2020-02-28
Fix spurious legacy-regex rule warning

selitvin:
The warning was occuring when both legacy and the new rule produced the
same overall set of fields, but maybe in different order. Although, both
rules would behave the same from the user perspective, they would see a
warning. The warning would be bogus because it wouldn't contain any
field names.

### DIFF
--- a/petastorm/unischema.py
+++ b/petastorm/unischema.py
@@ -421,13 +421,13 @@ def match_unischema_fields(schema, field_regex):
       expression patterns given by ``field_regex``.
     """
     if field_regex:
-        unischema_fields = []
-        legacy_unischema_fields = []
+        unischema_fields = set()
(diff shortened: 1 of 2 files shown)


petastorm/unischema.py
```diff
@@ -436,9 +436,9 @@ def match_unischema_fields(schema, field_regex):
             warnings.warn('schema_fields behavior has changed. Now, regular expression pattern must match'
                           ' the entire field name. The change in the behavior affects '
                           'the following fields: {}'.format(', '.join(diff_names)))
+        return list(unischema_fields)
     else:
-        unischema_fields = field_regex
-    return unischema_fields
+        return field_regex
```
fcchou: Nit: more clear to do `return []` here?

PR #503, opened 2020-03-10
Make petastorm reader support dataset url list.

WeichenXu123:
Currently, `make_batch_reader` and `make_reader` only accept a directory as a dataset url, but sometimes we need specify parquet file list as input, the reason is:

* We may want to use petastorm on aws S3, but S3 only provide eventually consistency for dir list operation (it is unreliable). So that, for S3 we need special code logic to get parquet file list and pass file list to  `make_batch_reader`

* Old version pyarrow do not allow files starts with underscore exists in the parquet directory (except the "_metadata" file). But some parquet implementation may put some commit protocol file such as _start, _succuss, make reader support dataset url list will address this issue.

## Test

UT added.

End to end test code:

~~~python

import numpy as np
import os

@pandas_udf('array<float>')
def gen_array(v):

df1 = spark.range(20).repartition(2).withColumn('v', gen_array('id'))

data_url = 'file:///tmp/t0001'
data_path = '/tmp/t0001'
df1.repartition(2).write.mode('overwrite').option("compression", "uncompressed").option("parquet.block.size", 1024 * 1024).parquet(data_url)

def get_pq_url_list(dir_path):
  lst = []
  for f in os.listdir(dir_path):
    if f.endswith('.parquet'):
      lst.append('file://' + os.path.join(dir_path, f))
  return lst

url_list = get_pq_url_list(data_path)
print(url_list)
reader = make_batch_reader(url_list)
for i in reader:
    print(str(i))

~~~

### DIFF
--- a/petastorm/fs_utils.py
+++ b/petastorm/fs_utils.py
@@ -20,6 +20,19 @@ from petastorm.gcsfs_helpers.gcsfs_wrapper import GCSFSWrapper
 from petastorm.hdfs.namenode import HdfsNamenodeResolver, HdfsConnector
 
 
+def get_dataset_path(parsed_url):
+    """
+    The dataset path is different than the one in `_parsed_dataset_url` for some filesystems.
+    For example s3fs expects the bucket name to be included in the path and doesn't support
+    paths that start with a `/`
+    """
+    if parsed_url.scheme.lower() in ['s3', 'gs', 'gcs']:
+        # s3fs expects paths of the form `bucket/path`
+        return parsed_url.netloc + parsed_url.path
+
+    return parsed_url.path
+
+
 class FilesystemResolver(object):
     """Resolves a dataset URL, makes a connection via pyarrow, and provides a filesystem object."""
 
@@ -158,12 +171,7 @@ class FilesystemResolver(object):
         For example s3fs expects the bucket name to be included in the path and doesn't support
         paths that start with a `/`
         """
-        if isinstance(self._filesystem, pyarrow.filesystem.S3FSWrapper) or \
-                isinstance(self._filesystem, GCSFSWrapper):
-            # s3fs expects paths of the form `bucket/path`
-            return self._parsed_dataset_url.netloc + self._parsed_dataset_url.path
-
-        return self._parsed_dataset_url.path
+        return get_dataset_path(self._parsed_dataset_url)
 
     def filesystem(self):
         """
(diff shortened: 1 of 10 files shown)


petastorm/fs_utils.py
```diff
@@ -183,3 +183,15 @@ def __getstate__(self):
         raise RuntimeError('Pickling FilesystemResolver is not supported as it may contain some '
                            'a file-system instance objects that do not support pickling but do not have '
                            'anti-pickling protection')
+
+
+def get_filesystem_and_path_or_paths(url_or_urls, hdfs_driver):
+    if isinstance(url_or_urls, list):
+        fs = FilesystemResolver(url_or_urls[0], hdfs_driver=hdfs_driver).filesystem()
+        path_or_paths = [FilesystemResolver(url, hdfs_driver=hdfs_driver).get_dataset_path()
```
selitvin: Can you please confirm that this does not end up being very expensive, especially in the case of `libhdfs` driver? (Recreating libhdfs filesystem, say 1000 times). 
Not too found of how the `FilesystemResolver` is design right now, perhaps an easy fix is to add an url-string only implementation of `get_dataset_path` as a standalone function, just based on the `urlparse().scheme`?

Can we easily raise an exception if the FS of all urls is not the same?

petastorm/reader.py
```diff
@@ -47,7 +47,16 @@
 _VENTILATE_EXTRA_ROWGROUPS = 2
 
 
-def make_reader(dataset_url,
+def _check_dataset_url(dataset_url):
```
selitvin: perhaps: `_check_dataset_url` -> `normalize_dataset_url`?

petastorm/reader.py
```diff
@@ -236,25 +241,22 @@ def make_batch_reader(dataset_url,
     :return: A :class:`Reader` object
     """
 
-    if dataset_url is None or not isinstance(dataset_url, six.string_types):
-        raise ValueError('dataset_url must be a string')
+    if isinstance(dataset_url_or_urls, list):
+        dataset_url_or_urls = [_check_dataset_url(url) for url in dataset_url_or_urls]
```
selitvin: Consider making `_check_dataset_url` to accept `dataset_url_or_urls` directly to avoid copy pasting this `if isinstance` code in multiple places.

petastorm/fs_utils.py
```diff
@@ -183,3 +183,15 @@ def __getstate__(self):
         raise RuntimeError('Pickling FilesystemResolver is not supported as it may contain some '
                            'a file-system instance objects that do not support pickling but do not have '
                            'anti-pickling protection')
+
+
+def get_filesystem_and_path_or_paths(url_or_urls, hdfs_driver):
```
selitvin: Please add a docstring.

WeichenXu123:
@selitvin 
`make_reader` test failed. Looks like the `make_reader` need more change to support reading url list.
But only support `make_batch_reader` is OK, because spark DL converter implements via `make_batch_reader`.

selitvin:
Having support for dataset url list only in `make_batch_reader` is ok with me if the `make_reader` adds too much work.

WeichenXu123:
@selitvin Ready.

petastorm/fs_utils.py
```diff
@@ -20,6 +20,19 @@
...
 
 
+def get_dataset_path(parsed_url):
+    """
+    The dataset path is different than the one in `_parsed_dataset_url` for some filesystems.
+    For example s3fs expects the bucket name to be included in the path and doesn't support
+    paths that start with a `/`
+    """
+    if parsed_url.scheme.lower() in ['s3', 'gs', 'gcs']:
+        # s3fs expects paths of the form `bucket/path`
```
selitvin: comment is slightly out of date (does not mention gs and gcs)

PR #504, opened 2020-03-12
Make `make_batch_reader` TransformSpec support output multi-dimensional array type.

WeichenXu123:
Make `make_batch_reader` TransformSpec support output multi-dimensional array type.

## Why we need this feature ?
The project: Simplify data conversion from Spark to TensorFlow: Spark converter basic implementation is implemented via `make_batch_reader` API. And user may want to do some preprocessing and return some tensor (multi-dimensional array) in the preprocess function.

Currently, `make_batch_reader` TransformSpec func only allow return one-dimensional array because of pyarrow format limitation.

## How does the PR address it ?
This PR address this issue. The approach is:

Flatten the multi-dimensional array returned by TransformSpec func, and in `ArrowReaderWorkerResultsQueueReader` loading data code, reshape back to the specified shape.
Note reshape is a in-place operation so it won't affect performance.

## Manual test code

~~~python

import os
import pandas as pd
import sys
import numpy as np

@pandas_udf('array<float>')
def gen_array(v):

df1 = spark.range(6).withColumn('v', gen_array('id'))

data_url = 'file:///tmp/t0001'
data_path = '/tmp/t0001'
df1.repartition(2).write.mode('overwrite').option("compression", "uncompressed").option("parquet.block.size", 1024 * 1024).parquet(data_url)

def preproc_fn(x):
  # reshape column 'v' to (2, 5) shape.
  x2 = pd.DataFrame({'v': x['v'].map(lambda x: x.reshape((2, 5))), 'id': x['id'] + 10000})
  return x2

spec = TransformSpec(
  preproc_fn,
  [('v', np.float32, (2, 5), False), ('id', np.int64, (), False)]
)

reader = make_batch_reader(data_url, num_epochs=1, transform_spec=spec)
for i in reader:
    print(i)
~~~

### DIFF
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -184,7 +184,7 @@ class ArrowReaderWorker(WorkerBase):
             transformed_schema_column_set = set([f.name for f in self._transformed_schema.fields.values()])
 
             if transformed_result_column_set != transformed_schema_column_set:
-                raise ValueError('Transformed result columns ({rs}) do not match required schema columns({sc})'
+                raise ValueError('Transformed result columns ({rc}) do not match required schema columns({sc})'
                                  .format(rc=','.join(transformed_result_column_set),
                                          sc=','.join(transformed_schema_column_set)))
 
@@ -201,7 +201,7 @@ class ArrowReaderWorker(WorkerBase):
             for field in self._transformed_schema.fields.values():
                 if len(field.shape) > 1:
                     transformed_result[field.name] = transformed_result[field.name] \
-                        .map(lambda x: check_shape_and_ravel(x, field))
+                        .map(lambda x: check_shape_and_ravel(x, field))  # pylint: disable=cell-var-from-loop
 
             result = pa.Table.from_pandas(transformed_result, preserve_index=False)
 
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -180,19 +180,28 @@ class ArrowReaderWorker(WorkerBase):
             for field_to_remove in set(transformed_result.columns) & set(self._transform_spec.removed_fields):
                 del transformed_result[field_to_remove]
 
+            transformed_result_column_set = set(transformed_result.columns)
+            transformed_schema_column_set = set([f.name for f in self._transformed_schema.fields.values()])
+
+            if transformed_result_column_set != transformed_schema_column_set:
+                raise ValueError('Transformed result columns ({rs}) do not match required schema columns({sc})'
+                                 .format(rc=','.join(transformed_result_column_set),
+                                         sc=','.join(transformed_schema_column_set)))
+
+            def check_shape_and_ravel(x, field):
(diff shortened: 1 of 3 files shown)


WeichenXu123:
Oh, sorry, it include some irrelative change. Let me update it.

selitvin:
Nice feature!

petastorm/arrow_reader_worker.py
```diff
@@ -174,6 +180,29 @@ def _load_rows(self, pq_file, piece, shuffle_row_drop_range):
...
 
+            transformed_result_column_set = set(transformed_result.columns)
+            transformed_schema_column_set = set([f.name for f in self._transformed_schema.fields.values()])
+
+            if transformed_result_column_set != transformed_schema_column_set:
+                raise ValueError('Transformed result columns ({rc}) do not match required schema columns({sc})'
+                                 .format(rc=','.join(transformed_result_column_set),
+                                         sc=','.join(transformed_schema_column_set)))
+
+            def check_shape_and_ravel(x, field):
```
selitvin: Probably a matter of personal taste, but maybe we take this out to be a private function? My scare is unintentional capturing of some outer-function state. It's also more testable when it's not an embedded function.

BTW, could be a good idea to unittest different branches in the function.

petastorm/arrow_reader_worker.py
```diff
@@ -174,6 +180,29 @@ def _load_rows(self, pq_file, piece, shuffle_row_drop_range):
...
+                    raise ValueError('field {name} must be the shape {shape}'
+                                     .format(name=field.name, shape=field.shape))
+                if not x.flags.c_contiguous:
+                    raise ValueError('Only support row major multi-dimensional array.')
+                return x.ravel()
+
+            for field in self._transformed_schema.fields.values():
+                if len(field.shape) > 1:
+                    transformed_result[field.name] = transformed_result[field.name] \
+                        .map(lambda x: check_shape_and_ravel(x, field))  # pylint: disable=cell-var-from-loop
```
selitvin: Instead of disabling the linter error, would this be safer?
```python
lambda x, f=field: check_shape_and_ravel(x, f)
```
Does the current code work for more than one field? Maybe extend the test to cover > 1 field scenario.

petastorm/arrow_reader_worker.py
```diff
@@ -174,6 +180,29 @@ def _load_rows(self, pq_file, piece, shuffle_row_drop_range):
...
+                if not isinstance(x, np.ndarray):
+                    raise ValueError('field {name} must be numpy array type.'.format(name=field.name))
+                if x.shape != field.shape:
+                    raise ValueError('field {name} must be the shape {shape}'
+                                     .format(name=field.name, shape=field.shape))
+                if not x.flags.c_contiguous:
+                    raise ValueError('Only support row major multi-dimensional array.')
+                return x.ravel()
+
+            for field in self._transformed_schema.fields.values():
```
selitvin: Can you please add a comment why we are raveling the fields?

WeichenXu123:
@selitvin Ready. :)

PR #506, opened 2020-03-16
Simplify data conversion from Spark to TensorFlow: support tensorflow dataset advance arguments

WeichenXu123:
## What changes are proposed in this PR?
Add `converter.make_tf_dataset()` with advanced params.

## API
~~~python
class SparkDatasetConverter(object):
    """
    A `SparkDatasetConverter` object holds one materialized spark dataframe and
    can be used to make one or more tensorflow datasets or torch dataloaders.
    The `SparkDatasetConverter` object is picklable and can be used in remote
    processes.
    See `make_spark_converter`
    """
    
    def make_tf_dataset(
            self,
            batch_size=32,
            prefetch=None,
            num_epochs=None,
            workers_count=4,
            **petastorm_reader_kwargs
    ):
        """
        Make a tensorflow dataset.

        This method will do the following two steps:
          1) Open a petastorm reader on the materialized dataset dir.
          2) Create a tensorflow dataset based on the reader created in (1)

        :param batch_size: The number of items to return per batch
        :param prefetch: Prefetch size for tensorflow dataset. If None will use
            tensorflow autotune size. Note only available on tensorflow>=1.14
        :param num_epochs: An epoch is a single pass over all rows in the dataset.
            Setting ``num_epochs`` to ``None`` will result in an infinite number
            of epochs.
        :param workers_count: An int for the number of workers to use in the
            reader pool. This only is used for the thread or process pool. Default value 4.
        :param petastorm_reader_kwargs: arguments for `petastorm.make_batch_reader()`,
            exclude these arguments: "dataset_url", "num_epochs", "workers_count".

        :return: a context manager for a `tf.data.Dataset` object.
                 when exit the returned context manager, the reader
                 will be closed.
        """
~~~

## Example code
~~~python
from petastorm import make_spark_converter
from petastorm.spark import SparkDatasetConverter
import torch

# specify a cache dir first.
# the dir is used to save materialized spark dataframe files
spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF, 'hdfs:/...')

df1 = ... # `df1` is a spark dataframe

# create a converter from `df1`
# it will materialize `df1` to cache dir.
converter1 = make_spark_converter(df1)

# make a tensorflow dataset from `converter1
with converter1.make_tf_dataset(
        batch_size=128, prefetch=2, num_epochs=1, worker_count=1,
        cur_shard=hvd.rank(),
        shard_count=hvd.count()) as dataset:
    # the `dataset` is `tf.data.Dataset` object
    # we can train/evaluate model on `dataset`
    # when exit the with context, the underlying petastorm reader of the dataset will be closed
    ...

# delete the cached files of the dataframe.
converter1.delete()
~~~

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -26,12 +26,13 @@ from pyarrow import LocalFileSystem
 from pyspark.sql.session import SparkSession
 from six.moves.urllib.parse import urlparse
 
-import petastorm
+from petastorm import make_batch_reader
 from petastorm.fs_utils import FilesystemResolver
 
 DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
 
 logger = logging.getLogger(__name__)
+
 
 def _get_spark_session():
     return SparkSession.builder.getOrCreate()
@@ -243,7 +244,7 @@ class TFDatasetContextManager(object):
         from petastorm.tf_utils import make_petastorm_dataset
         import tensorflow as tf
 
-        self.reader = petastorm.make_batch_reader(self.data_url, **self.petastorm_reader_kwargs)
+        self.reader = make_batch_reader(self.data_url, **self.petastorm_reader_kwargs)
 
         dataset = make_petastorm_dataset(self.reader) \
             .flat_map(tf.data.Dataset.from_tensor_slices)
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -116,7 +116,7 @@ def _delete_cache_data_atexit(dataset_url):
         warnings.warn('delete cache data {url} failed.'.format(url=dataset_url))
 
 
-def get_env_rank_and_size():
+def _get_env_rank_and_size():
     """
     Get rank and size from environment, return (rank, size), if failed, return (None, None)
     """
@@ -134,18 +134,15 @@ def get_env_rank_and_size():
     return None, None
 
 
-def is_rank_and_size_consistent_with_env(petastorm_reader_kwargs):
+def _is_rank_and_size_consistent_with_horovod(cur_shard, shard_count, hvd_rank, hvd_size):
     """
-    Check whether the cur_shard and shard_count args are consistent with environment variables.
-    If not consistent with environment variables, return False.
-    If there're no related environment variable set, return True.
+    Check whether the cur_shard and shard_count args are consistent with horovod environment variables.
+    If not consistent with horovod environment variables, return False.
+    If there're no related horovod environment variable set, return True.
     """
-    env_rank, env_size = get_env_rank_and_size()
-    if env_rank is not None and env_size is not None:
-        if petastorm_reader_kwargs.get('cur_shard') != env_rank or \
-                petastorm_reader_kwargs.get('shard_count') != env_size:
+    if hvd_rank is not None and hvd_size is not None:
+        if cur_shard != hvd_rank or shard_count != hvd_size:
             return False
-
     return True
 
 
@@ -178,7 +175,7 @@ class SparkDatasetConverter(object):
 
     def make_tf_dataset(
             self,
-            batch_size=32,
+            batch_size=None,
             prefetch=None,
             num_epochs=None,
             workers_count=None,
@@ -191,7 +188,9 @@ class SparkDatasetConverter(object):
           1) Open a petastorm reader on the materialized dataset dir.
           2) Create a tensorflow dataset based on the reader created in (1)
 
-        :param batch_size: The number of items to return per batch
+        :param batch_size: The number of items to return per batch. Default None.
+            If None, current implementation will set batch size to be 32, in future,
+            None value will denotes auto tuned best value for batch size.
         :param prefetch: Prefetch size for tensorflow dataset. If None will use
             tensorflow autotune size. Note only available on tensorflow>=1.14
         :param num_epochs: An epoch is a single pass over all rows in the dataset.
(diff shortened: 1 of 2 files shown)


petastorm/spark/spark_dataset_converter.py
```diff
@@ -112,6 +116,21 @@ def _delete_cache_data_atexit(dataset_url):
         warnings.warn('delete cache data {url} failed.'.format(url=dataset_url))
 
 
+def get_env_rank_and_size():
```
mengxr: docstring

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +158,55 @@ def __len__(self):
         """
         return self.dataset_size
 
-    def make_tf_dataset(self):
+    def make_tf_dataset(
+            self,
+            batch_size=32,
+            prefetch=None,
+            num_epochs=None,
+            workers_count=4,
```
mengxr: change to None to reserve autotune in the future. mention current default in the docstring.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +158,55 @@ def __len__(self):
...
                  when exit the returned context manager, the reader
                  will be closed.
         """
-        return TFDatasetContextManager(self.cache_dir_url)
+
+        # override some arguments default values of petastorm reader
+        petastorm_reader_kwargs['num_epochs'] = num_epochs
+        petastorm_reader_kwargs['workers_count'] = workers_count
+
+        env_rank, env_size = get_env_rank_and_size()
```
mengxr: * move it to a utility function since it will be used by both TF and torch
* `env_rank` -> rank

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +158,55 @@ def __len__(self):
...
+        if env_rank is not None and env_size is not None:
+            if petastorm_reader_kwargs['cur_shard'] != env_rank or \
```
mengxr: handle key error

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +158,55 @@ def __len__(self):
...
+                    petastorm_reader_kwargs['shard_count'] != env_size:
+                logger.warning('The petastorm arguments cur_shard and shard_count is not '
+                               'consistent with detected MPI/horovod environments, do you '
```
mengxr: it is okay to let each work sample the whole dataset in different order

petastorm/spark/spark_dataset_converter.py
```diff
@@ -166,17 +221,44 @@ class TFDatasetContextManager(object):
...
-        self.dataset = make_petastorm_dataset(self.reader)
+        self.reader = make_batch_reader(self.data_url, **self.petastorm_reader_kwargs)
 
-    def __enter__(self):
-        return self.dataset
+        dataset = make_petastorm_dataset(self.reader) \
+            .flat_map(tf.data.Dataset.from_tensor_slices)
+        dataset = dataset.batch(batch_size=self.batch_size)
+
+        if LooseVersion(tf.__version__) >= LooseVersion('1.14'):
```
mengxr: 1.14 is not the first release to have prefetch

petastorm/spark/spark_dataset_converter.py
```diff
@@ -166,17 +221,44 @@ class TFDatasetContextManager(object):
...
+            # We can make prefetch optimization
+            prefetch = self.prefetch
+            if prefetch is None:
+                prefetch = tf.data.experimental.AUTOTUNE
+            if prefetch != 0:
+                dataset = dataset.prefetch(prefetch)
```
mengxr: not needed if prefetch() handles 0 value gracefully.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +176,58 @@ def __len__(self):
         """
         return self.dataset_size
 
-    def make_tf_dataset(self):
+    def make_tf_dataset(
+            self,
+            batch_size=32,
```
selitvin: Do not have a very strong opinion here, but I would expect user to explicitly set the batch size - 32 seems too arbitrary.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -112,6 +116,39 @@ def _delete_cache_data_atexit(dataset_url):
         warnings.warn('delete cache data {url} failed.'.format(url=dataset_url))
 
 
+def get_env_rank_and_size():
```
selitvin: Let's prefix with `_` as both `get_env_rank_and_size` and `is_rank_and_size_consistent_with_env` are private.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -139,19 +176,58 @@ def __len__(self):
...
+
+        # override some arguments default values of petastorm reader
+        petastorm_reader_kwargs['num_epochs'] = num_epochs
+        if workers_count is None:
+            # TODO: generate a best tuned value for default worker count value
+            workers_count = 4
+        petastorm_reader_kwargs['workers_count'] = workers_count
+
+        if not is_rank_and_size_consistent_with_env(petastorm_reader_kwargs):
+            logger.warning('The petastorm arguments cur_shard and shard_count is not '
```
selitvin: Find it hard to understand this message. Can we add more information here? Perhaps show current petastorm sharding settings and horovod ones and suggest a course of action to fix the issue.

Do you mean `each work sample` -> `each worker sample`?

If it does sample whole dataset, is that a problem?

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -237,3 +239,57 @@ def test_change_cache_dir_raise_error(test_ctx):
     # restore conf (other test need use it)
     test_ctx.spark.conf.set('petastorm.spark.converter.parentCacheDirUrl', test_ctx.temp_url)
     assert test_ctx.temp_url == _get_parent_cache_dir_url()
+
```
selitvin: Can you please add a test that covers horovod rank compatibility check?

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -237,3 +239,57 @@ def test_change_cache_dir_raise_error(test_ctx):
...
+        return make_batch_reader(dataset_url, **kwargs)
+
+    petastorm.spark.spark_dataset_converter.make_batch_reader = mock_fn
+    try:
+        yield captured_args
+    finally:
+        petastorm.spark.spark_dataset_converter.make_batch_reader = make_batch_reader
+
+
+def test_tf_dataset_petastorm_args(test_ctx):
```
selitvin: Is there a reason not to use  `mock.patch` here? there won't be need in a custom `mock_make_batch_reader` implementation.

petastorm/spark/spark_dataset_converter.py
mengxr: How about changing it to None but keep the internal default to 32. So we can automatically decide the best batch size in the future? If users do not like it, they can always set a value.

WeichenXu123:
@selitvin Ready.
Note: the `unittest.mock` has bug on python3 (some version) which make CI raise error. So I import built-in mock first and if failed fallback to unittest mock.
~~~python
try:
  from mock import mock
except:
  from unitest import mock
~~~

PR #511, opened 2020-03-21
[WIP][ML-10118] Keep petastorm dataset/dataloader schema fields order the same with spark dataframe

WeichenXu123:
[WIP]Keep petastorm dataset/dataloader schema fields order the same with spark dataframe.

### DIFF
--- a/petastorm/transform.py
+++ b/petastorm/transform.py
@@ -17,7 +17,7 @@ from petastorm.unischema import UnischemaField, Unischema
 
 
 class TransformSpec(object):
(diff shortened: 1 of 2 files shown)


WeichenXu123:
@mengxr 
Because there's this PR https://github.com/uber/petastorm/pull/512
we don't need add a `selected_field` into TransformSpec. We can directly infer schema from transformed result(and schema field order will be the field order of the transformed result pandas dataframe).
Create a new PR https://github.com/uber/petastorm/pull/513

PR #512, opened 2020-03-23
[WIP] Auto infer schema (including fields shape) from the first row

WeichenXu123:
## What issues does the PR addresses ?

There're 2 issues in `make_batch_reader`, one is critical and another is less critical but a pain point.

### (Critical) Inferring schema in `make_batch_reader` cannot infer fields' shape information
Because there's no shape information, when make tensorflow dataset from the reader, if we make some tensorflow dataset operations, such as unroll, batch, and reshape field, error may occur. Tensorflow graph operator depends on field shape information heavily.

### (Pain point) The `TransformSpec` need to specify edit/removed fields manually
We hope user can only provide a transform function, and petastorm can automatically infer the result schema from the output pandas dataframe of the transform function.

## The approach in the PR
Add a method `ArrowReaderWorker. infer_schema_from_first_row` which can read a row first and infer the schema from the row. So that we can infer the accurate shape information.
Add a param `infer_schema_from_first_row` into `make_batch_reader` (default off, so won't break API behavior)

**Limitations:**
* for all rows (before applying predicates), require all values in each field non-nullable and having the same shape.

## Test
Unit test to be added. But it is ready for first review.

## Example code

~~~python
import os
import pandas as pd
import sys
import numpy as np
import tensorflow as tf

from petastorm.spark import make_spark_converter
spark.conf.set('petastorm.spark.converter.parentCacheDirUrl', 'file:/tmp/converter')

data_url = 'file:/tmp/0001'
data_path = '/tmp/t0001'

@pandas_udf('array<float>')
def gen_array(v):

df1 = spark.range(10).withColumn('v', gen_array('id')).repartition(2)
cv1 = make_spark_converter(df1)

# we can auto infer one-dim array shape
with cv1.make_tf_dataset(batch_size=4, num_epochs=1) as dataset:
	iter = dataset.make_one_shot_iterator()
	next_op = iter.get_next()
	with tf.Session() as sess:
		for i in range(3):
			batch = sess.run(next_op)
			print(batch)


def preproc_fn(x):
  # reshape column 'v' to (2, 5) shape.
  x2 = pd.DataFrame({'v': x['v'].map(lambda x: x.reshape((2, 5))), 'id': x['id'] + 10000})
  return x2

# now we can auto infer multi-dim array shape.
with cv1.make_tf_dataset(batch_size=4, preprocess_fn=preproc_fn, num_epochs=1) as dataset:
	iter = dataset.make_one_shot_iterator()
	next_op = iter.get_next()
	with tf.Session() as sess:
		for i in range(3):
			batch = sess.run(next_op)
			print(batch)
~~~

### DIFF
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -170,7 +170,7 @@ class ArrowReaderWorker(WorkerBase):
         if all_cols:
             self.publish_func(all_cols)
 
-    def infer_schema_from_first_row(self):
+    def infer_schema_from_a_row(self):
         self._init_dataset()
 
         piece0 = self._split_pieces[0]
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -24,6 +24,7 @@ from pyarrow.parquet import ParquetFile
 
 from petastorm.cache import NullCache
 from petastorm.compat import compat_piece_read, compat_table_columns_gen, compat_column_data
+from petastorm.unischema import Unischema, UnischemaField
 from petastorm.workers_pool import EmptyResultError
 from petastorm.workers_pool.worker_base import WorkerBase
 
@@ -111,6 +112,18 @@ class ArrowReaderWorker(WorkerBase):
     def new_results_queue_reader():
         return ArrowReaderWorkerResultsQueueReader()
 
+    def _init_dataset(self):
+        if not self._dataset:
+            self._dataset = pq.ParquetDataset(
+                self._dataset_path_or_paths,
+                filesystem=self._filesystem,
+                validate_schema=False)
+
+        if self._dataset.partitions is None:
+            # When read from parquet file list, the `dataset.partitions` will be None.
+            # But other petastorm code require at least an empty `ParquetPartitions` object.
+            self._dataset.partitions = pq.ParquetPartitions()
+
     # pylint: disable=arguments-differ
     def process(self, piece_index, worker_predicate, shuffle_row_drop_partition):
         """Main worker function. Loads and returns all rows matching the predicate from a rowgroup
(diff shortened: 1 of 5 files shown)


petastorm/arrow_reader_worker.py
```diff
@@ -168,6 +170,38 @@ def process(self, piece_index, worker_predicate, shuffle_row_drop_partition):
         if all_cols:
             self.publish_func(all_cols)
 
+    def infer_schema_from_first_row(self):
```
liangz1: nit: I'm not sure whether the partition[0] necessarily contains the "first" row? Could the partitions be out of order? If so, we may call it `infer_schema_from_a_row`.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -216,6 +230,18 @@ def make_tf_dataset(
             workers_count = 4
         petastorm_reader_kwargs['workers_count'] = workers_count
 
+        if 'dataset_url_or_urls' in petastorm_reader_kwargs:
+            raise ValueError('User cannot set dataset_url_or_urls argument.')
+
+        if 'transform_spec' in petastorm_reader_kwargs or \
+                'infer_schema_from_first_row' in petastorm_reader_kwargs:
+            raise ValueError('User cannot set transform_spec and infer_schema_from_first_row '
```
liangz1: Shall we also allow users to use `transform_spec`&`infer_schema_from_first_row`?  Keeping `transform_spec` would make it consistent with the rest of the petastorm library.

petastorm/arrow_reader_worker.py
WeichenXu123: Here I read the first row in the index-0 row-groups. But index-0 row-groups may be non-deterministic ? Not sure. `infer_schema_from_a_row` sounds good.

petastorm/spark/spark_dataset_converter.py
WeichenXu123: I think the param `preprocess_fn` should cover the functionality of `transform_spec`, and it is easier to use (can auto inferring result schema), so I forbid the two params.

WeichenXu123:
I create a simple PR to address issue 1, https://github.com/uber/petastorm/pull/517
We can merge that one first.
This PR could be a long-term work.

PR #513, opened 2020-03-23
[ML-10118] Preserve spark dataframe schema order when create petastorm dataset/dataloader

WeichenXu123:
* Preserve spark dataframe schema order when create petastorm dataset/dataloader
* Add selected fields for `TransformSpec`

### DIFF
--- a/petastorm/transform.py
+++ b/petastorm/transform.py
@@ -70,12 +70,11 @@ def transform_schema(schema, transform_spec):
                                                 shape=field_to_edit[2], codec=None, nullable=field_to_edit[3])
         fields.append(edited_unischema_field)
 
-    unknown_field_names = set(transform_spec.selected_fields) - set(f.name for f in fields)
-    if unknown_field_names:
-        warnings.warn('selected_fields specified some field names that are not part of the schema. '
-                      'These field names will be ignored "{}". '.format(', '.join(unknown_field_names)))
-
     if transform_spec.selected_fields is not None:
+        unknown_field_names = set(transform_spec.selected_fields) - set(f.name for f in fields)
+        if unknown_field_names:
+            warnings.warn('selected_fields specified some field names that are not part of the schema. '
+                          'These field names will be ignored "{}". '.format(', '.join(unknown_field_names)))
         fields = [f for f in fields if f.name in transform_spec.selected_fields]
         fields = sorted(fields, key=lambda f: transform_spec.selected_fields.index(f.name))
 
--- a/petastorm/transform.py
+++ b/petastorm/transform.py
@@ -42,7 +42,7 @@ class TransformSpec(object):
         self.func = func
         self.edit_fields = edit_fields or []
 
-        if self.removed_fields is not None and self.selected_fields is not None:
+        if removed_fields is not None and selected_fields is not None:
             raise ValueError('User can only specify one of removed_fields and selected_fields in TransformSpec.')
 
         self.removed_fields = removed_fields or []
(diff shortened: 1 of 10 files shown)


petastorm/unischema.py
```diff
@@ -95,11 +100,14 @@ def get(parent_schema_name, field_names):
         :return: A namedtuple with field names defined by `field_names`
         """
         # Cache key is a combination of schema name and all field names
-        sorted_names = list(sorted(field_names))
```
liangz1: I think we just need to remove this sort operation here and no need to check _UNISCHEMA_FIELD_ORDER, because this `get()` function is only called in unischema.py#L243, and param `field_names` is already an `odict_keys` object, and the _UNISCHEMA_FIELD_ORDER checking should already be done in the `Unischema.__init__()`.

petastorm/transform.py
```diff
@@ -34,10 +34,18 @@ def __init__(self, func=None, edit_fields=None, removed_fields=None):
...
+          If selected_fields specified, The reader schema will preserve the field order in selected_fields.
+
+        Note: For param `removed_fields` and `selected_fields`, user can only specify one of them.
         """
         self.func = func
         self.edit_fields = edit_fields or []
         self.removed_fields = removed_fields or []
+        self.selected_fields = selected_fields or []
+
+        if self.removed_fields and self.selected_fields:
```
liangz1: If `removed_fields` overlaps with `edit_fields`, the original behavior here is to edit the common fields (ignore `removed_fields`). We can also keep the behavior consistent for `selected_fields`: just don't raise an error when `removed_fields` is also provided.

petastorm/unischema.py
WeichenXu123: Yes. That's right. But, for safety, we'd better keep the old behavior of the internal method `_NamedtupleCache.get()` when _UNISCHEMA_FIELD_ORDER set 'alphabeta'. The method `_NamedtupleCache.get` is possible to be used in other places.

petastorm/transform.py
WeichenXu123: Yes. I thought this. But, I think force user providing only one of them make the semantic simpler and leading less error.

petastorm/transform.py
```diff
@@ -61,4 +69,8 @@ def transform_schema(schema, transform_spec):
                                                 shape=field_to_edit[2], codec=None, nullable=field_to_edit[3])
         fields.append(edited_unischema_field)
 
+    if transform_spec.selected_fields:
+        fields = [f for f in fields if f.name in transform_spec.selected_fields]
+        fields = sorted(fields, key=lambda f: transform_spec.selected_fields.index(f.name))
```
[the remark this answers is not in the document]
WeichenXu123: Add a warning, similar to removed_fields

petastorm/unischema.py
```diff
@@ -31,6 +31,11 @@
 
 from petastorm.compat import compat_get_metadata
 
+# _UNISCHEMA_FIELD_ORDER available values are 'preserve_input_order' or 'alphabetical'
+# default behavior is 'preserve_input_order', but, when backward compatibility required,
+# set this to be 'alphabetical'
```
selitvin: If we are deprecating the alphabetical let's state it here explicitly.

petastorm/unischema.py
selitvin: I like the idea that we keep the old behavior behind the `_UNISCHEMA_FIELD_ORDER`, but would vote for deprecating the old behavior in a couple of versions so we don't need to maintain the old behavior forever.

liangz1: Sounds good!

PR #520, opened 2020-03-25
Added is_petastorm_compatible to make_reader

abditag2:
This allows make_reader to read datasets that are compatible with petastorm but not generated with it. For instance, a dataset with primary types such as ints and floats arrays.


selitvin:
@abditag2, do we want to land this?

abditag2:
@evgeny-goldin It is not needed anymore. Given that make_batch_reader is faster for the target dataset, It is probably better to raise the exception so the users use the faster kind of reader. I will close this. 

I HorovodEstimators, I ended up using both types of readers. If there is a transform to happen, I use make_reader otherwise, I use make_batch_reader.

PR #525, opened 2020-03-30
[ML-10160] Refine Spark dataset converter warning logs.

WeichenXu123:
* Warn users if the intermediate file sizes are too small (<50mb)
* Warn users if set wrong parent cache dir. (Only in databricks environment)

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -425,7 +425,7 @@ def _check_parent_cache_dir_url(dir_url):
     if 'DATABRICKS_RUNTIME_VERSION' in os.environ and not _is_spark_local_mode():
         if isinstance(fs, LocalFileSystem):
             # User need to use dbfs fuse URL.
-            if dir_path.startswith('/dbfs/'):
+            if not dir_path.startswith('/dbfs/'):
                 raise ValueError(
                     "You must specify a dbfs fuse path for {conf}, like: 'file:/dbfs/path/to/cache_dir'"
                     .format(conf=SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF))
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -543,7 +543,7 @@ def _check_dataset_file_median_size(url_list):
         mid_index = len(file_size_list) // 2
         median_size = sorted(file_size_list)[mid_index]
         if median_size < 50 * 1024 * 1024:
-            logger.warning('The median size ({%d}) of these parquet files ({%s}) is too small.'
+            logger.warning('The median size (%d) of these parquet files (%s) is too small.'
                            'Increase file sizes by repartition or coalesce spark dataframe, which '
                            'will help improve performance.', median_size, ','.join(url_list))
 
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -18,16 +18,18 @@ import logging
 import os
 import shutil
 import threading
+import time
 import uuid
 import warnings
 from distutils.version import LooseVersion
+from multiprocessing.pool import ThreadPool
 
 from pyarrow import LocalFileSystem
 from pyspark.sql.session import SparkSession
 from six.moves.urllib.parse import urlparse
 
 from petastorm import make_batch_reader
-from petastorm.fs_utils import FilesystemResolver
+from petastorm.fs_utils import FilesystemResolver, get_filesystem_and_path_or_paths
 
 DEFAULT_ROW_GROUP_SIZE_BYTES = 32 * 1024 * 1024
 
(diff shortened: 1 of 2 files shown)


petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -42,11 +45,11 @@
 from petastorm.spark.spark_dataset_converter import (
     _check_url, _get_horovod_rank_and_size, _get_parent_cache_dir_url,
     _check_rank_and_size_consistent_with_horovod, _make_sub_dir_url,
-    register_delete_dir_handler, _wait_file_available)
+    register_delete_dir_handler, _wait_file_available, _check_dataset_file_median_size,
```
liangz1: nit: I usually use`isort` (a tool installed from pip) to sort the imports in a source file so that it follows the PEP 8 standard.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -412,6 +416,20 @@ def _check_url(dir_url):
             'Please prepend "file://" for local filesystem.'.format(dir_url))
 
 
+def _check_parent_cache_dir_url(dir_url):
+    """
+    Check dir url whether is suitable to be used as parent cache directory.
+    """
+    _check_url(dir_url)
```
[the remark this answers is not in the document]
WeichenXu123: other places also use `_check_url`

petastorm/spark/spark_dataset_converter.py
```diff
@@ -517,6 +535,19 @@ def wait_for_file(path):
...
 
 
+def _check_dataset_file_median_size(url_list):
+    fs, path_list = get_filesystem_and_path_or_paths(url_list)
+    # TODO: also check other file system.
+    if isinstance(fs, LocalFileSystem):
+        file_size_list = [os.path.getsize(path) for path in path_list]
+        mid_index = len(file_size_list) // 2
+        median_size = sorted(file_size_list)[mid_index]
+        if median_size < 50 * 1024 * 1024:
```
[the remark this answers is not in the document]
WeichenXu123: number of files being small should not be an issue. Because when reading, each file will be split into row-groups, one row-group is a minimal reading task.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -517,6 +536,19 @@ def wait_for_file(path):
         pool.join()
 
 
+def _check_dataset_file_median_size(url_list):
+    fs, path_list = get_filesystem_and_path_or_paths(url_list)
+    # TODO: also check other file system.
+    if isinstance(fs, LocalFileSystem):
```
selitvin: Would it make sense to check also for non-local file systems? Not sure how fast does it work with fuse - would it make sense to acquire file sizes on a threadpool?

petastorm/spark/spark_dataset_converter.py
```diff
@@ -517,6 +536,19 @@ def wait_for_file(path):
         pool.join()
 
 
+def _check_dataset_file_median_size(url_list):
```
selitvin: Where is the function being called from?

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -478,3 +494,36 @@ def delay_create_file2():
...
+        '/a/b/01.parquet': 50,
+        '/a/b/02.parquet': 70,
+        '/a/b/03.parquet': 60,
+        '/a/b/04.parquet': 65,
+        '/a/b/05.parquet': 999000,
+    }
+    with mock.patch('os.path.getsize') as mock_path_get_size:
+        mock_path_get_size.side_effect = lambda p: file_size_map[p]
+        url_list = ['file://' + path for path in file_size_map.keys()]
+        with patch_logger('petastorm.spark.spark_dataset_converter') as output:
```
selitvin: pytest appears to have a [built in log capture](https://docs.pytest.org/en/latest/logging.html). Do you think we should use it instead of a custom one?

petastorm/spark/spark_dataset_converter.py
WeichenXu123: * check also for non-local file systems make sense, but pyarrow filesystem interface do not implement the `stat` method. It need additional work so we could do it later.
* acquire file sizes by a threadpool make sense.

petastorm/spark/spark_dataset_converter.py
WeichenXu123: Oh, forgot, will call it when dataframe materialized finished.

petastorm/tests/test_spark_dataset_converter.py
WeichenXu123: Oh sure if the builtin one works fine.

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -61,6 +64,19 @@ def tear_down(self):
         self.spark.stop()
 
 
+@contextlib.contextmanager
+def patch_logger(name):
```
selitvin, line 68: Unused.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -517,6 +536,25 @@ def wait_for_file(path):
...
+    fs, path_list = get_filesystem_and_path_or_paths(url_list)
+
+    # TODO: also check file size for other file system.
+    if isinstance(fs, LocalFileSystem):
+        pool = ThreadPool(64)
+        try:
+            file_size_list = pool.map(os.path.getsize, path_list)
+            mid_index = len(file_size_list) // 2
+            median_size = sorted(file_size_list)[mid_index]
```
mengxr, line 548: minor: use median from statistics module

petastorm/spark/spark_dataset_converter.py
```diff
@@ -412,6 +416,21 @@ def _check_url(dir_url):
...
+    """
+    Check dir url whether is suitable to be used as parent cache directory.
+    """
+    _check_url(dir_url)
+    fs, dir_path = get_filesystem_and_path_or_paths(dir_url)
+    if 'DATABRICKS_RUNTIME_VERSION' in os.environ and not _is_spark_local_mode():
+        if isinstance(fs, LocalFileSystem):
+            # User need to use dbfs fuse URL.
+            if not dir_path.startswith('/dbfs/'):
+                raise ValueError(
```
mengxr, line 429: Can we make it a warning instead? User might have NFS mount, e.g., using EFS.

petastorm/spark/spark_dataset_converter.py
WeichenXu123: python2 don't have `statistics` package so need to install it by pip. Do we need it ?

PR #531, opened 2020-04-04
Fix issue: spark session created on spark executor

WeichenXu123:
Current code exists an issue that spark session will be created on spark executor.
This is because `_get_spark_session` being invoked by `_wait_file_available`, and `_wait_file_available` will be invoked from executor side.

So I fix it. I removed the config `petastorm.spark.converter.fileAvailabilityWaitTimeoutSecs`, the config is rarely used.

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -562,7 +562,7 @@ _FILE_AVAILABILITY_WAIT_TIMEOUT_SECS = 30
 
 def _wait_file_available(url_list):
     """
-    Waiting about SparkDatasetConverter.FILE_AVAILABILITY_WAIT_TIMEOUT_SECS_CONF seconds to make sure
+    Waiting about _FILE_AVAILABILITY_WAIT_TIMEOUT_SECS seconds (default 30 seconds) to make sure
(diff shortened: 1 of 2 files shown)


petastorm/spark/spark_dataset_converter.py
```diff
@@ -551,22 +557,20 @@ def _materialize_df(df, parent_cache_dir_url, parquet_row_group_size_bytes,
     return save_to_dir_url
 
 
+_FILE_AVAILABILITY_WAIT_TIMEOUT_SECS = 30
+
+
 def _wait_file_available(url_list):
     """
     Waiting about SparkDatasetConverter.FILE_AVAILABILITY_WAIT_TIMEOUT_SECS_CONF seconds to make sure
```
liangz1, line 565: This comment is outdated.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -45,6 +45,13 @@ def vector_to_array(_1, _2='float32'):
 
 
 def _get_spark_session():
+    """
+    Get or create spark session.
+    Note: This function can only be invoked from driver side.
+    """
+    if pyspark.TaskContext.get() is not None:
```
selitvin, line 52: Can we add a test to verify the condition?

PR #534, opened 2020-04-07
[ML-10366]  Fix bug: cleanup metadata in converter.delete()

WeichenXu123:
When the user wants to manually delete the cache files and create a new cache, they do:

~~~python
converter.delete()
converter = make_spark_converter(df)
~~~
Get error:
```
org.apache.spark.sql.AnalysisException: Path does not exist: file:/dbfs/ml/tmp/petastorm/QA/keras/20200407043623-appid-app-20200406162107-0000-d7d7cdfd-3739-48d0-9dc1-bffba08fff12;
```

Fix: We should cleanup cache metadata in SparkDatasetConverter.delete() besides deleting the data files.

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -106,16 +106,9 @@ def register_delete_dir_handler(handler):
         _delete_dir_handler = handler
 
 
-def _delete_cache_data(dataset_url):
-    """
-    Delete the cache data in the underlying file system.
-    """
-    _delete_dir_handler(dataset_url)
-
-
 def _delete_cache_data_atexit(dataset_url):
     try:
-        _delete_cache_data(dataset_url)
+        _delete_dir_handler(dataset_url)
     except Exception:  # pylint: disable=broad-except
         logger.warning('delete cache data %s failed.', dataset_url)
 
@@ -281,7 +274,7 @@ class SparkDatasetConverter(object):
         """
         Delete cache files at self.cache_dir_url.
         """
-        _delete_cache_data(self.cache_dir_url)
+        _remove_df_from_cache(self.cache_dir_url)
 
 
 class TFDatasetContextManager(object):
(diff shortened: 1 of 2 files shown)


petastorm/spark/spark_dataset_converter.py
```diff
@@ -475,13 +468,22 @@ def _cache_df_or_retrieve_cache_data_url(df, parent_cache_dir_url,
...
         # do not find cached dataframe, start materializing.
-        cached_df_meta = CachedDataFrameMeta.create_cached_dataframe(
+        cached_df_meta = CachedDataFrameMeta.create_cached_dataframe_meta(
             df, parent_cache_dir_url, parquet_row_group_size_bytes,
             compression_codec, dtype)
         _cache_df_meta_list.append(cached_df_meta)
         return cached_df_meta.cache_dir_url
 
 
+def _remove_df_from_cache(cache_dir_url):
```
liangz1, line 478: What about we call it `_remove_cache_metadata_and_data(...)` and rename `_delete_dir_handler` as `_remove_cache_data_handler`? It would be less confusing which function is doing what, and also match the semantics of `delete_cache_data_atexit` (basically we call the underlying files "data" and the in-memory structure "metadata").

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -234,6 +234,20 @@ def test_df_caching(test_ctx):
     assert converter12.cache_dir_url != converter22.cache_dir_url
 
 
+def test_df_delete_caching_meta(test_ctx):
+    from petastorm.spark.spark_dataset_converter import _cache_df_meta_list
+    df1 = test_ctx.spark.range(10)
+    df2 = test_ctx.spark.range(20)
+    converter1 = make_spark_converter(df1)
+    converter2 = make_spark_converter(df2)
+    converter1.delete()
```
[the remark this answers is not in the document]
WeichenXu123: I think we letting `make_tf_dataset` raise 'dir not exist' error will be good enough.
It is hard to make an ideal destruction, because multiple converters may share the same cache dir, destructing all of them makes code complex.

petastorm/spark/spark_dataset_converter.py
WeichenXu123: `_remove_cache_metadata_and_data` is good.
keep `_delete_dir_handler` for it can be used in other places when deleting dir.

PR #537, opened 2020-04-09
Fix bug: respect dynamically changed parent cache dir conf

liangz1:
Bug: If we execute the following code:
```python
spark.conf.set("petastorm.spark.converter.parentCacheDirUrl", "file:///url1")
converter1 = make_spark_converter(df)   # This works fine.
# Change conf
spark.conf.set("petastorm.spark.converter.parentCacheDirUrl", "file:///url2")
converter2 = make_spark_converter(df)
```
The last line will hit cache:
```The median size (1333465) of these parquet files (file:/url1/file_abc.parquet) is too small.Increase file sizes by repartition or coalesce spark dataframe, which will help improve performance.```
indicating that the new conf parent dir is not respected (still hitting the .../url1 dir).

Fix: We respect the conf change by adding an equality test against the parent cache dir.

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -384,7 +384,7 @@ def _get_df_plan(df):
 
 class CachedDataFrameMeta(object):
 
-    def __init__(self, df, row_group_size, compression_codec, dtype):
+    def __init__(self, df, parent_cache_dir_url, row_group_size, compression_codec, dtype):
         self.row_group_size = row_group_size
         self.compression_codec = compression_codec
         # Note: the metadata will hold dataframe plan, but it won't
@@ -394,11 +394,12 @@ class CachedDataFrameMeta(object):
         self.df_plan = _get_df_plan(df)
         self.cache_dir_url = None
         self.dtype = dtype
+        self.parent_cache_dir_url = parent_cache_dir_url
 
     @classmethod
     def create_cached_dataframe(cls, df, parent_cache_dir_url, row_group_size,
                                 compression_codec, dtype):
-        meta = cls(df, row_group_size, compression_codec, dtype)
+        meta = cls(df, parent_cache_dir_url, row_group_size, compression_codec, dtype)
         meta.cache_dir_url = _materialize_df(
             df,
             parent_cache_dir_url=parent_cache_dir_url,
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: petastorm/spark/spark_dataset_converter.py [hunk 1]
```diff
@@ -478,7 +479,8 @@ def _cache_df_or_retrieve_cache_data_url(df, parent_cache_dir_url,
             if meta.row_group_size == parquet_row_group_size_bytes and \
                     meta.compression_codec == compression_codec and \
                     meta.df_plan.sameResult(df_plan) and \
-                    meta.dtype == dtype:
+                    meta.dtype == dtype and \
+                    meta.parent_cache_dir_url == parent_cache_dir_url:
```
WeichenXu123, line 483: Before comparing, let's apply `petastorm.reader.normalize_dataset_url` first.

liangz1: I added the comparison after getting the conf so that all the saved parent_cache_dir is already normalized.

petastorm/tests/test_spark_dataset_converter.py
```diff
@@ -236,6 +236,17 @@ def test_df_caching(test_ctx):
...
 
+    ori_temp_url = test_ctx.spark.conf.get(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF)
+    tempdir = tempfile.mkdtemp('_spark_converter_test1')
+    new_temp_url = 'file://' + tempdir.replace(os.sep, '/')
+    test_ctx.spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF,
+                            new_temp_url)
+    assert ori_temp_url != new_temp_url
+    converter13 = make_spark_converter(df1)
+    assert converter1.cache_dir_url != converter13.cache_dir_url
+    test_ctx.spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF,
```
WeichenXu123, line 247: * Let's add a test for path with or without trailing '/' (they should be regarded as the same parent dir)

* Let's move the line restoring conf `test_ctx.spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF,` into finally block, so that when test raise error, it won't influence other tests.

WeichenXu123:
Nit: `normalize_dataset_url` apply on a parent dir, this make code confusing, so rename `normalize_dataset_url` to `normalize_dir_url` and move it into `petastorm.fs_utils` package.

PR #538, opened 2020-04-09
Fix: The "median size too small" warning is too frequent

liangz1:
Problem: This warning below shows up even when there is only one partition, and the file list could be too long to read.
```
The median size (1333465) of these parquet files (file list...) is too small.Increase file sizes by repartition or coalesce spark dataframe, which will help improve performance.
```

Fix: 
1. Only print warning for (num_of_files > 1 && median file size < 50MB)
2. Remove the file list in the warning.
3. Provide the total size of the dataset and the threshold file size (50MB) in the warning to help users to decide how many partitions do they need.

Test:
Added unit test in `test_check_dataset_file_median_size`.

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -594,28 +594,21 @@ def _wait_file_available(url_list):
 
 def _check_dataset_file_median_size(url_list):
     fs, path_list = get_filesystem_and_path_or_paths(url_list)
-    SIZE_LIMIT_BYTES = 50 * 1024 * 1024
-    MB_BYTES = 1024 * 1024
+    RECOMMENDED_FILE_SIZE_BYTES = 50 * 1024 * 1024
 
     # TODO: also check file size for other file system.
     if isinstance(fs, LocalFileSystem):
         pool = ThreadPool(64)
         try:
             file_size_list = pool.map(os.path.getsize, path_list)
-            total_size = sum(file_size_list)
-            num_partitions = len(file_size_list)
-            if total_size > SIZE_LIMIT_BYTES:
-                mid_index = num_partitions // 2
-                median_size = sorted(file_size_list)[mid_index]
-                if median_size < SIZE_LIMIT_BYTES:
-                    logger.warning('The median size %d MB (< 50 MB) of the parquet files is too small. '
-                                   'Total size: %d MB. Increase the median file size by calling df.repartition(n) or '
-                                   'df.coalesce(n), which will help improve the performance. Parquet files: %s, ...',
-                                   median_size // MB_BYTES, total_size // MB_BYTES, url_list[0])
-            else:
-                if num_partitions > 8:
-                    logger.warning("The total size of the parquet files is small (%d MB) enough to fit in "
-                                   "a single node, consider using df.coalesce(1) to improve the performance.")
+            if len(file_size_list) > 1:
+                mid_index = len(file_size_list) // 2
+                median_size = sorted(file_size_list, reverse=True)[mid_index]  # take the larger one if tie
+                if median_size < RECOMMENDED_FILE_SIZE_BYTES:
+                    logger.warning('The median size %d B (< 50 MB) of the parquet files is too small. '
+                                   'Total size: %d B. Increase the median file size by calling df.repartition(n) or '
+                                   'df.coalesce(n), which might help improve the performance. Parquet files: %s, ...',
+                                   median_size, sum(file_size_list), url_list[0])
         finally:
             pool.close()
             pool.join()
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -603,14 +603,19 @@ def _check_dataset_file_median_size(url_list):
         try:
             file_size_list = pool.map(os.path.getsize, path_list)
             total_size = sum(file_size_list)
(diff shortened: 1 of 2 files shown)


petastorm/spark/spark_dataset_converter.py
```diff
@@ -592,18 +592,23 @@ def wait_for_file(path):
...
-            median_size = sorted(file_size_list)[mid_index]
-            if median_size < 50 * 1024 * 1024:
-                logger.warning('The median size (%d) of these parquet files (%s) is too small.'
-                               'Increase file sizes by repartition or coalesce spark dataframe, which '
-                               'will help improve performance.', median_size, ','.join(url_list))
+            total_size = sum(file_size_list)
+            if total_size > SIZE_LIMIT_BYTES:
+                mid_index = len(file_size_list) // 2
+                median_size = sorted(file_size_list)[mid_index]
+                if median_size < SIZE_LIMIT_BYTES:
```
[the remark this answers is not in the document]
liangz1: Making up 3 different thresholds seems pretty complex and arbitrary... Maybe we just keep it simple for now and only use `SIZE_LIMIT_BYTES`. It could also be easier for users to figure out how to get rid of this message: by reducing the numPartitions as much as possible until the median size reaches 50 MB.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -592,18 +592,23 @@ def wait_for_file(path):
...
+                    logger.warning('The median size %d MB (< 50 MB) of the parquet files is too small. '
+                                   'Total size: %d MB. Increase the median file size by calling df.repartition(n) or '
+                                   'df.coalesce(n), which will help improve the performance.',
+                                   median_size // MB_BYTES, total_size // MB_BYTES)
```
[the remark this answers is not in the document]
liangz1: Sure, I can add the dir URL to the end of the message so that the long URL would not stretch the sentence too much.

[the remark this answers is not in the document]
WeichenXu123: add warning for total_size < SIZE_LIMIT_BYTES and num_files > 100 ?

petastorm/spark/spark_dataset_converter.py
```diff
@@ -594,18 +594,28 @@ def wait_for_file(path):
...
         try:
             file_size_list = pool.map(os.path.getsize, path_list)
-            mid_index = len(file_size_list) // 2
-            median_size = sorted(file_size_list)[mid_index]
-            if median_size < 50 * 1024 * 1024:
-                logger.warning('The median size (%d) of these parquet files (%s) is too small.'
-                               'Increase file sizes by repartition or coalesce spark dataframe, which '
-                               'will help improve performance.', median_size, ','.join(url_list))
+            total_size = sum(file_size_list)
+            num_partitions = len(file_size_list)
```
mengxr, line 606: Do not call it num_partitions. It is not equivalent to number of files.

petastorm/spark/spark_dataset_converter.py
```diff
@@ -594,18 +594,28 @@ def wait_for_file(path):
...
+            if total_size > SIZE_LIMIT_BYTES:
```
mengxr, line 607: Shall we make the logic simpler:

~~~python
if len(path_lists) > 1:
 median_size = sorted(file_size_list, reverse=True)[mid_index]  # take the larger one if tie
 if median_size < RECOMMEND_FILE_SIZE_BYTES:
   logger.warning('...')
# no else branch
~~~

petastorm/spark/spark_dataset_converter.py
```diff
@@ -594,18 +594,28 @@ def wait_for_file(path):
...
+                mid_index = num_partitions // 2
+                median_size = sorted(file_size_list)[mid_index]
+                if median_size < SIZE_LIMIT_BYTES:
+                    logger.warning('The median size %d MB (< 50 MB) of the parquet files is too small. '
+                                   'Total size: %d MB. Increase the median file size by calling df.repartition(n) or '
+                                   'df.coalesce(n), which will help improve the performance. Parquet files: %s, ...',
+                                   median_size // MB_BYTES, total_size // MB_BYTES, url_list[0])
```
mengxr, line 614: * `//` would produce zero. It is fine to show number of bytes.
* "will help" -> "might help".

petastorm/spark/spark_dataset_converter.py
```diff
@@ -594,18 +594,21 @@ def wait_for_file(path):
...
             file_size_list = pool.map(os.path.getsize, path_list)
-            mid_index = len(file_size_list) // 2
-            median_size = sorted(file_size_list)[mid_index]
-            if median_size < 50 * 1024 * 1024:
-                logger.warning('The median size (%d) of these parquet files (%s) is too small.'
-                               'Increase file sizes by repartition or coalesce spark dataframe, which '
-                               'will help improve performance.', median_size, ','.join(url_list))
+            if len(file_size_list) > 1:
+                mid_index = len(file_size_list) // 2
+                median_size = sorted(file_size_list, reverse=True)[mid_index]  # take the larger one if tie
```
liangz1, line 606: We should remove `reverse=True` in order to "take the larger one if tie".

mengxr: thanks for the catch! mind submitting a fix?

liangz1: sg

PR #542, opened 2020-04-15
Address several autograph failed issues for TF2

WeichenXu123:
Currently, for some function, TF2 autograph will fail.
See
https://github.com/tensorflow/tensorflow/issues/35765
https://github.com/tensorflow/tensorflow/issues/30149
https://github.com/tensorflow/autograph/issues/3

If autograph failed, the functions will be run eagerly and TF cannot optimize them. So we'd better address them.

## Manually test
~~~python
df1 = spark.range(100)
from petastorm.spark import make_spark_converter

# Set a cache directory on DBFS FUSE for intermediate data.
spark.conf.set("petastorm.spark.converter.parentCacheDirUrl", "file:///dbfs/ml/tmp/petastorm/QA/bugs/")
converter1 = make_spark_converter(df1)

with converter1.make_tf_dataset(num_epochs=1) as dataset:
  for batch in dataset:
    print(batch.id)
~~~

* Before
Output includes:
```
WARNING:tensorflow:AutoGraph could not transform <function _NamedtupleCache.get at 0x7f0bfbe6f200> and will run it as-is.
Please report this to the TensorFlow team. When filing the bug, set the verbosity to 10 (on Linux, `export AUTOGRAPH_VERBOSITY=10`) and attach the full output.
Cause: expected exactly one node node, found [<gast.gast.FunctionDef object at 0x7f0bfad5d050>, <gast.gast.Return object at 0x7f0bfad5d7d0>]
WARNING:tensorflow:AutoGraph could not transform <function make_petastorm_dataset.<locals>.<lambda> at 0x7f0bf8da03b0> and will run it as-is.
Cause: could not parse the source code:

            .map(lambda row: _set_shape_to_named_tuple(reader.schema, row, reader.batched_output))

This error may be avoided by creating the lambda in a standalone statement.
```

* After
The warnings listed above disappear.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,12 +82,14 @@ script:
     --ignore=examples/mnist/tests/test_pytorch_mnist.py
     --ignore=petastorm/tests/test_pytorch_utils.py
     --ignore=petastorm/tests/test_pytorch_dataloader.py
+    --ignore=petastorm/tests/test_tf_autograph.py
     petastorm examples
 
   - $RUN $PYTEST -m "not forked" -v -Y  --cov-append
       --ignore=examples/mnist/tests/test_pytorch_mnist.py
       --ignore=petastorm/tests/test_pytorch_utils.py
       --ignore=petastorm/tests/test_pytorch_dataloader.py
+      --ignore=petastorm/tests/test_tf_autograph.py
       petastorm examples
 
     # Tensorflow and pytorch tests confict (segfault). Split into separate runs.
@@ -95,6 +97,9 @@ script:
       examples/mnist/tests/test_pytorch_mnist.py
       petastorm/tests/test_pytorch_dataloader.py
       petastorm/tests/test_pytorch_utils.py
+
+    # Run test_tf_autograph separately because the error message "AutoGraph could not transform..." only show once, so we need ensure no other TF tests run before it.
+  - $RUN $PYTEST petastorm/tests/test_tf_autograph.py
 
 after_success:
   - codecov --required
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,7 +85,7 @@ script:
     --ignore=petastorm/tests/test_tf_autograph.py
     petastorm examples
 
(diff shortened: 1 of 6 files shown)


selitvin:
How did you find these failures? Just by running in your external environment? If so, is it hard to add a test to make sure we don't break autograph going forward?

petastorm/unischema.py
```diff
@@ -239,7 +239,7 @@ def create_schema_view(self, fields):
 
         return Unischema('{}_view'.format(self._name), view_fields)
 
-    def _get_namedtuple(self):
```
mengxr, line 242: do not rename it is not needed for this PR

petastorm/tests/test_tf_autograph.py
```diff
@@ -0,0 +1,33 @@
...
+    spark_ctx = TestContext()
+    caplog.clear()
+    df1 = spark_ctx.spark.range(100)
+    converter1 = make_spark_converter(df1)
+    results = []
+    with converter1.make_tf_dataset(num_epochs=1) as dataset:
+        for batch in dataset:
+            results.append(batch)
+    assert "AutoGraph could not transform" not in " ".join(caplog.messages)
+    spark_ctx.tear_down()
```
selitvin, line 33: Let's either make `TestContext` into a contextmanager or call tear_down in a `finally` clause.

.travis.yml
```diff
@@ -96,6 +98,9 @@ script:
       petastorm/tests/test_pytorch_dataloader.py
       petastorm/tests/test_pytorch_utils.py
 
+    # Run test_tf_autograph separately because the error message "AutoGraph could not transform..." only show once, so we need ensure no other TF tests run before it.
+  - $RUN $PYTEST petastorm/tests/test_tf_autograph.py
```
selitvin, line 102: Please add `-v` (actually, can move `-v` into PYTEST= definition), `-Y` and `--cov-append` to this invocation (this is the reason, I think, for all-red codecov).

WeichenXu123: -v was already in PYTEST definition

PR #544, opened 2020-04-16
Fix DataLoader iter(dataloader) cannot be called more than once

WeichenXu123:
### Issue
PyTorch DataLoader allows iter(datalaoder) to be called more than once. Specifically, the user can define a function:
```
def train_one_epoch(dataloader, ...):
    for batch in dataloader:
        ...
```
and use the function with the same dataloader:
```
with DataLoader(...) as dataloader:
    for epoch in range(num_epochs):
        train_one_epoch(dataloader, ...)
```
In the current implementation, when `for batch in dataloader` is called in the second epoch, no batch will be returned.

This PR will enable the usage above.

### Fix
When each new iteration starts: reset iterator internal status: including
* create new _shuffling_buffer
* reset petastorm reader

And add a limit, only after we finish the last iteration, we can start a new iteration because these iterators share the same underlying reader, so they cannot run in parallel.


petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -175,3 +175,13 @@ def test_with_batch_reader(scalar_dataset, shuffling_queue_capacity):
         if pa.__version__ != '0.15.0':
             assert len(scalar_dataset.data) == sum(batch['int_fixed_size_list'].shape[0] for batch in batches)
             assert batches[0]['int_fixed_size_list'].shape[1] == len(scalar_dataset.data[0]['int_fixed_size_list'])
+
+
+@pytest.mark.timeout(100)
```
liangz1, line 180: This line is actually not needed.

petastorm/pytorch.py
```diff
@@ -150,6 +144,22 @@ def __iter__(self):
 
         keys = None
 
+        if self._in_iter:
+            raise RuntimeError('Only after previous iteration finished we can start another iteration.')
```
mengxr, line 148: minor: This is user-facing error message but it doesn't provide any instructions. How about:

~~~
You must finish a full pass of Petastorm DataLoader before making another pass from the beginning.
If you do need to terminate early and restart from beginning, please re-create the reader and the data loader.
~~~

petastorm/pytorch.py
```diff
@@ -131,15 +131,9 @@ def __init__(self, reader, batch_size=1, collate_fn=decimal_friendly_collate,
...
-            # and give up on the unbound growth protection mechanism.
-            min_after_dequeue = shuffling_queue_capacity - 1
-            self._shuffling_buffer = RandomShufflingBuffer(shuffling_queue_capacity,
-                                                           min_after_retrieve=min_after_dequeue,
-                                                           extra_capacity=100000000)
-        else:
-            self._shuffling_buffer = NoopShufflingBuffer()
+        self.shuffling_queue_capacity = shuffling_queue_capacity
+        self._in_iter = False
+        self._iter_pass = 0
```
mengxr, line 136: Do we need to remember the number of passes? We can set `_in_iter` to (None, False, True) to tell whether it is the first pass.

petastorm/pytorch.py
```diff
@@ -192,6 +202,9 @@ def __iter__(self):
         if self._batch_acc:
             yield self.collate_fn(self._batch_acc)
 
+        self._in_iter = False
```
mengxr, line 205: Could you add an inline comment saying that we purposely didn't not wrap it under a final block?

petastorm/pytorch.py
```diff
@@ -150,6 +144,22 @@ def __iter__(self):
 
         keys = None
 
+        if self._in_iter:
+            raise RuntimeError('Only after previous iteration finished we can start another iteration.')
+        self._in_iter = True
+        if self._iter_pass > 0:
+            self.reader.reset()
```
mengxr, line 151: Log a warning message saying that reseting reader to position 0.

petastorm/pytorch.py
```diff
@@ -150,6 +148,23 @@ def __iter__(self):
 
         keys = None
 
+        if self._in_iter:
+            raise RuntimeError(_PARALLEL_ITER_ERROR)
+        if self._in_iter is not None:
+            self.reader.reset()
+            logging.warn('Start a new pass of Petastorm DataLoader, reset underlying Petastorm reader to position 0.')
```
selitvin, line 155: Let's use 
```python
logger = logging.getLogger(__name__)
logger.warn(..)
```
this will give our users better granularity in the control over log levels of different parts of their system.

petastorm/pytorch.py
```diff
@@ -277,6 +284,28 @@ def __iter__(self):
 
         keys = None
 
+        if self._in_iter:
```
selitvin, line 287: nit: maybe it can make the code clearer, if we are more explicit about the fact that this is a three state and not just a boolean, i.e. `if self._in_iter is not None and self._in_iter == True:`?

petastorm/pytorch.py
```diff
@@ -277,6 +284,28 @@ def __iter__(self):
...
+            raise RuntimeError(_PARALLEL_ITER_ERROR)
+        if self._in_iter is not None:
+            self.reader.reset()
+            logging.warn('Start a new pass of Petastorm DataLoader, reset underlying Petastorm reader to position 0.')
```
selitvin, line 291: `warn` is deprecated. Please use `warning`.

petastorm/pytorch.py
```diff
@@ -308,6 +337,10 @@ def __iter__(self):
         for batch in self._yield_batches(keys):
             yield batch
 
+        # We purposely do not wrap it under a final block, because must finish a full pass of Petastorm DataLoader
+        # before making another pass from the beginning
+        self._in_iter = False
```
selitvin, line 342: What would happen if the consumer of batches raises an error? Perhaps would be safer to put `self._in_iter` into a `finally` clause? (would be nice to test this case)

petastorm/pytorch.py
```diff
@@ -92,7 +94,39 @@ def decimal_friendly_collate(batch):
...
+            raise RuntimeError(_PARALLEL_ITER_ERROR)
+        if self._in_iter is not None:
+            self.reader.reset()
+            logger.warning('Start a new pass of Petastorm DataLoader, reset underlying Petastorm reader to position 0.')
+        self._in_iter = True
+
+        try:
+            for batch in self._iter_impl():
+                yield batch
+        except RuntimeError as e:
```
selitvin, line 122: Curious, why did you choose `RuntimeError` and not higher exception class (`Exception`)?

petastorm/pytorch.py
```diff
@@ -92,7 +94,39 @@ def decimal_friendly_collate(batch):
...
+            self._error = e
```
selitvin, line 123: Should we `logger.error` more information here?

petastorm/pytorch.py
```diff
@@ -92,7 +94,39 @@ def decimal_friendly_collate(batch):
...
+            raise e
```
selitvin, line 124: `raise e` would erase inner exception call-stack information. Consider `raise` without arguments instead.

petastorm/pytorch.py
```diff
@@ -92,7 +94,40 @@ def decimal_friendly_collate(batch):
...
+
+    def __iter__(self):
+        if self._error is not None:
+            raise RuntimeError('Cannot start a new iteration because last time iteration failed with error {err}.'
+                               .format(err=repr(self._error)))
+        if self._in_iter is not None and self._in_iter == True:  # noqa: E712
+            raise RuntimeError(_PARALLEL_ITER_ERROR)
+        if self._in_iter is not None:
+            self.reader.reset()
+            logger.warning('Start a new pass of Petastorm DataLoader, reset underlying Petastorm reader to position 0.')
```
selitvin, line 116:
```suggestion
            logger.info('Start a new pass of Petastorm DataLoader, reset underlying Petastorm reader to position 0.')
```

PR #549, opened 2020-04-21
Import petastorm.spark in init

praateekmahajan:
## Issue

1. I install petastorm `pip install petastorm`
2. Run `from petastorm.spark import X, Y, Z` and I get the error 
```python
ModuleNotFoundError : No module named 'petastorm.spark'
```
## Fix
- Added import petastorm.spark in init
- Updated the version so pip builds happen.

I'm **not sure if this is the right way**, since there are nuances about releases and how petastorm currently does python module management.

### DIFF
--- a/petastorm/__init__.py
+++ b/petastorm/__init__.py
@@ -17,8 +17,9 @@ import warnings
 from petastorm.errors import NoDataAvailableError  # noqa: F401
 from petastorm.reader import make_reader, make_batch_reader  # noqa: F401
 from petastorm.transform import TransformSpec  # noqa: F401
+import petastorm.spark
 
-__version__ = '0.8.2'
+__version__ = '0.8.3'
 
 if sys.version_info.major < 3:
     warnings.warn('Petastorm on Python 2 is deprecated and will remove Python 2 support in next release.')


praateekmahajan:
Thanks @selitvin that makes sense. The only reason I created this PR, I thought it's faster than an issue, as it shows the requested code change. However that might have not been a good decision, so I can move it to an issue if you want.

I updated my PR to reflect what the issue is, which is 
1. I install petastorm `pip install petastorm`
2. Run `from petastorm.spark import X, Y, Z` and I get the error 
```python
ModuleNotFoundError : No module named 'petastorm.spark'
```

selitvin:
Petastorm does not have a spark subpackage (there is new petastorm.spark package in the next release, but not in the existing public one). What were you trying to import?

praateekmahajan:
Oh I see, the subpackage/release part makes sense. I can close the PR and wait out the next release. Out of curiosity where are subpackages defined?

I was trying to do the following (from the [example](https://github.com/uber/petastorm/blob/master/examples/spark_dataset_converter/pytorch_converter_example.py))

`from petastorm.spark import SparkDatasetConverter, make_spark_converter`.

Abhishaike:
Hey @praateekmahajan, running  `import petastorm.spark` also gives me the error

`ModuleNotFoundError : No module named 'petastorm.spark'`

Not quite understanding your fix, any advice?

PR #555, opened 2020-05-20
In-memory cache

abditag2:
This PR adds a in-memory cache to the Petastorm loader. This enables very fast reading of data and removing the IO/netowrk bottleneck.


fps7806:
Can you explain what exactly is being cached? From a quick glance I couldn't figure out the difference between the two dataloaders.

abditag2:
@fps7806 Caching works by preserving the values loaded into the ShufflingBuffer into the memory or GPU and not removing them. This way, if in-memory-cache is enabled, the values will only be read once from disk/network into memory. Each worker only caches its own shard of the data and not the entire data set.

petastorm/tests/test_parquet_reader.py
```diff
@@ -202,3 +189,16 @@ def preproc_fn1(x):
         assert len(sample) == 2
         assert (2, 3) == sample['tensor_col_1'].shape[1:] and \
             (3, 4, 5) == sample['tensor_col_2'].shape[1:]
+
+
+@pytest.mark.parametrize('reader_factory', _D)
```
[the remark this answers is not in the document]
selitvin: Changing the order should have not fixed it. Not for this PR, but wonder what's going on here..

selitvin (commented):
Please remove accidentally added `petastorm/hdfs/tests/__pycache__/test_hdfs_namenode.cpython-36-pytest-6.1.2.pyc.17`

petastorm/tests/test_shuffling_buffer.py
```diff
@@ -203,8 +301,10 @@ def test_noop_shuffling_buffer_stream_through(buffer_type):
 def test_batched_random_shuffling_buffer_stream_through(batch_size):
     """Feed a 0:99 sequence through a BatchedRandomShufflingBuffer. Check that the order has changed."""
     input_sequence = range(100)
-    a = _feed_a_sequence_through_the_queue(BatchedRandomShufflingBuffer(10, 3, batch_size), input_sequence)
-    b = _feed_a_sequence_through_the_queue(BatchedRandomShufflingBuffer(10, 3, batch_size), input_sequence)
+    a = _feed_a_sequence_through_the_queue(
```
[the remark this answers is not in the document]
abditag2: In the case of batch_size of 10, `_feed_a_sequence_through_the_queue` falls into a forever loop where it can neither add or retrieve samples from the buffer. It is not a but, simply, those numbers are a weird combination.

petastorm/pytorch.py
```diff
@@ -303,20 +322,33 @@ def _iter_impl(self):
...
             min_after_dequeue = self.shuffling_queue_capacity - 1
             shuffling_queue_capacity = min_after_dequeue + self.batch_size
-            self._shuffling_buffer = BatchedRandomShufflingBuffer(
-                shuffling_queue_capacity,
-                min_after_retrieve=min_after_dequeue,
-                extra_capacity=100000000,
-                batch_size=self.batch_size
-            )
+
+            def _instantiate_bateched_shuffling_buffer():
```
selitvin, line 334:
```suggestion
            def _instantiate_batched_shuffling_buffer():
```

petastorm/pytorch.py
```diff
@@ -339,16 +372,33 @@ def _iter_impl(self):
...
         for batch in self._yield_batches(keys):
+            if self.inmemory_cache_all:
+                self._other_shuffling_buffer.add_many(batch.values())
             yield batch
 
+        if self.inmemory_cache_all:
+            for _ in range(self.num_epochs - 1 if self.num_epochs else sys.maxsize):
+                self._other_shuffling_buffer.finish()
+                self._shuffling_buffer = self._other_shuffling_buffer
+                self._other_shuffling_buffer = instantiate_buffer_fn()
```
selitvin, line 396: Can self._other_shuffling_buffer be just local variables and not part of object state (we should probably also make self._shuffling_buffer a local variable, but it was like this before, so should not change in this PR)?

petastorm/pytorch.py
```diff
@@ -303,20 +323,33 @@ def _iter_impl(self):
         # the requested batch_size ready.
 
         keys = None
+
+        instantiate_buffer_fn = None
```
selitvin, line 327: Maybe we refactor this large if into a separate function?

petastorm/pytorch.py
```diff
@@ -339,16 +372,33 @@ def _iter_impl(self):
...
+                self._other_shuffling_buffer.add_many(batch.values())
             yield batch
 
+        if self.inmemory_cache_all:
+            for _ in range(self.num_epochs - 1 if self.num_epochs else sys.maxsize):
+                self._other_shuffling_buffer.finish()
+                self._shuffling_buffer = self._other_shuffling_buffer
+                self._other_shuffling_buffer = instantiate_buffer_fn()
+                for batch in self._yield_batches(keys):
+                    if self.inmemory_cache_all:
```
selitvin, line 398: This condition must be true here due to if on line 392. That being said, we could optimize and not populate the buffer in the last epoch.

petastorm/pytorch.py
```diff
@@ -283,7 +287,11 @@ def __init__(self, reader, batch_size=1, transform_fn=None,
         :param batch_size: the number of items to return per batch; factored into the len() of this reader
         :param transform_fn: an optional callable to convert batches from the reader to PyTorch tensors
         :param shuffling_queue_capacity: Queue capacity is passed to the underlying :class:`tf.RandomShuffleQueue`
-          instance. If set to 0, no suffling will be done.
+          instance. If set to 0, no shuffling will be done.
+        :param inmemory_cache_all: This is an boolean that indicates whether the data should be
```
selitvin, line 291: Maybe we should mention that the user should expect batch size smaller then requested on the epoch boundary.

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -208,3 +209,153 @@ def test_call_iter_on_dataloader_multiple_times(synthetic_dataset, reader_factor
...
+            BatchedDataLoader(reader, num_epochs=2)
+
+
+@pytest.mark.parametrize('cache_type', ['mem_cache', None])
+@pytest.mark.parametrize('shuffling_queue_capacity', [20, 0])
+@pytest.mark.parametrize('reader_factory', [make_batch_reader, make_reader])
+def test_in_memory_cache_one_epoch(two_columns_non_petastorm_dataset, cache_type,
+                                   shuffling_queue_capacity, reader_factory):
+    batch_size = 10
+    print("testing reader_factor: {}, cache_type: {}, shuffling_queue_capacity: {}"
```
selitvin, line 236: Please remove debug printouts

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -208,3 +209,153 @@ def test_call_iter_on_dataloader_multiple_times(synthetic_dataset, reader_factor
...
+
+def test_mem_cache_num_epochs_without_mem_cache_error(two_columns_non_petastorm_dataset):
+    error_string = "num_epochs should not be specified when inmemory_cache_all is not enabled."
+    with make_batch_reader(two_columns_non_petastorm_dataset.url,
+                           num_epochs=1) as reader:
+        with pytest.raises(ValueError, match=error_string):
+            BatchedDataLoader(reader, num_epochs=2)
+
+
+@pytest.mark.parametrize('cache_type', ['mem_cache', None])
```
selitvin, line 230:
```suggestion
@pytest.mark.parametrize('inmemory_cache_all', [True, False])
```

selitvin (approved):

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -208,3 +209,99 @@ def test_call_iter_on_dataloader_multiple_times(synthetic_dataset, reader_factor
...
+@pytest.mark.parametrize('shuffling_queue_capacity', [20, 0])
+@pytest.mark.parametrize('reader_factory', [make_batch_reader, make_reader])
+@pytest.mark.parametrize('num_epochs', [1, 2, None])
```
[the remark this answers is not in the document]
abditag2: added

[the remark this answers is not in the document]
abditag2: Added

PR #563, opened 2020-06-24
Use mypy in our CI script

selitvin:
Using mypy to run static type checking as part of travis build

Fixed some issues causing mypy to fail:
- Upgraded minimal torch and Pillow versions in `setup.py`
- Upgraded torch and Pillow in `.travis.yml`. This is a quick fix. A proper better way is to upgrade docker image used for the build.
- Rely on `unittest`'s `mock`
- Adding some type hints required by mypy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,8 @@ before_script:
 
   - $RUN pip list
 
+  - $RUN mypy petastorm
+
   # We run a bunch of linting in before_script to fail fast and not run unit tests in case of lint failure.
   - $RUN flake8 . --count --show-source --statistics
   - $RUN flake8 . --count --exit-zero --max-complexity=10 --statistics
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,5 @@
+[mypy]
+ignore_missing_imports = True
+
+[mypy-petatstorm.unischema.*]
+ignore_errors = True
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ REQUIRED_PACKAGES = [
     'dill>=0.2.1',
     'diskcache>=3.0.0',
     'future>=0.10.2',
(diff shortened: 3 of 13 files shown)


mypy.ini
```diff
@@ -0,0 +1,5 @@
+[mypy]
+ignore_missing_imports = True
```
jsgoller1, line 2: Do we want this to be true? Is there a particular case where this causes some kind of false positive?

.travis.yml
```diff
@@ -57,10 +57,15 @@ before_script:
   - $RUN pip install --upgrade numpy
 
   - $RUN pip install -U pyarrow==${PYARROW_VERSION} tensorflow==${TF_VERSION}
+
+  # TODO(yevgeni): need to update docker build image at some stage
+  - $RUN pip install -U torchvision==0.2.1 torch==1.5.1 Pillow==7.1.2
```
jsgoller1, line 62: Can we do this [in the Dockerfile](https://github.com/uber/petastorm/blob/master/docker/Dockerfile#L34)? Should be a really quick one-line change, I think. Possibly saves us a nasty headache later on if trying to debug package issues.

mypy.ini
selitvin: I tried listing external libraries one-by-one when they failed mypy for not good reasons. Gave up eventually.

.travis.yml
selitvin: Done. Going to land the PR.

setup.py
```diff
@@ -33,6 +33,7 @@
     'dill>=0.2.1',
     'diskcache>=3.0.0',
     'future>=0.10.2',
+    'gcsfs>=0.2.0',
```
mengxr, line 36: @selitvin Is `gcsfs` required?

PR #564, opened 2020-07-02
exposed pyarrow filters in the make_reader and make_batch_reader api

abditag2:


petastorm/reader.py
```diff
@@ -67,7 +67,8 @@ def make_reader(dataset_url,
                 cache_type='null', cache_location=None, cache_size_limit=None,
                 cache_row_size_estimate=None, cache_extra_settings=None,
                 hdfs_driver='libhdfs3',
-                transform_spec=None):
+                transform_spec=None,
+                pyarrow_filters=None):
```
selitvin, line 71: Please add to the docstring. Don't know if a good idea, but you may update README.rst as it seems like a pretty useful option.

petastorm/reader.py
```diff
@@ -187,7 +189,8 @@ def make_batch_reader(dataset_url_or_urls,
                       cache_type='null', cache_location=None, cache_size_limit=None,
                       cache_row_size_estimate=None, cache_extra_settings=None,
                       hdfs_driver='libhdfs3',
-                      transform_spec=None):
+                      transform_spec=None,
+                      pyarrow_filters=None):
```
selitvin, line 193: Please add to the docstring.

petastorm/arrow_reader_worker.py
```diff
@@ -99,6 +99,7 @@ def __init__(self, worker_id, publish_func, args):
         self._local_cache = args[5]
         self._transform_spec = args[6]
         self._transformed_schema = args[7]
+        self._pyarrow_filters = args[8]
```
jsgoller1, line 102: @selitvin - not super familiar with this part of the codebase but is there any reason here why we're using positional args like this instead of keyword args in the `__init__()` or `**kwargs`?

selitvin: Historical reasons. Feel free to clean up :)

petastorm/reader.py
abditag2: added

petastorm/reader.py
abditag2: I added to the table comparing make_reader and make_batch_reader.

petastorm/arrow_reader_worker.py
abditag2: Cleaned up!

.travis.yml
```diff
@@ -57,7 +57,7 @@ before_script:
   - $RUN pip install --upgrade numpy
 
   - $RUN pip install -U pyarrow==${PYARROW_VERSION} tensorflow==${TF_VERSION}
-  - if [[ ${PYSPARK_VERSION} > "3" ]] ; then $RUN pip install https://ml-team-public-read.s3-us-west-2.amazonaws.com/pyspark-3.1.0.dev0-60dd1a690fed62b1d6442cdc8cf3f89ef4304d5a.tar.gz --force-reinstall --upgrade ; fi
+  - $RUN pip install pyspark==${PYSPARK_VERSION} --force-reinstall --upgrade
```
abditag2, line 60: @selitvin Please review this part to make sure this is ok.

petastorm/tests/test_end_to_end.py
```diff
@@ -857,3 +862,32 @@ def test_make_batch_reader_with_url_list(scalar_dataset):
...
 
         assert row_count == 100
+
+
+def test_pyarrow_filters_make_reader(synthetic_dataset):
+    with make_reader(synthetic_dataset.url, workers_count=5, num_epochs=1,
+                     filters=[('partition_key', '=', 'p_5'), ]) as reader:
+        uv = set()
+        for data in reader:
+            uv.add(data[0])
```
selitvin, line 872: Can we use an explicit namedtuple name here instead of `data[0]` for readability?

petastorm/tests/test_end_to_end.py
```diff
@@ -857,3 +862,32 @@ def test_make_batch_reader_with_url_list(scalar_dataset):
...
+
+        assert uv == {'p_5'}
+
+
+def test_pyarrow_filters_make_batch_reader():
+    path = tempfile.mkdtemp()
+    url = 'file://' + path
+
+    data = create_test_scalar_dataset(url, 100, partition_by=['datetime'], two_datetime_vals=True)
+    SyntheticDataset = namedtuple('synthetic_dataset', ['url', 'data', 'path'])
```
selitvin, line 882: Not sure, why do we need an additional named tuple here? Can we just use `make_batch_reader(data.url, ...)`?

petastorm/tests/test_common.py
```diff
@@ -159,7 +159,8 @@ def _partition_key_to_str(row):
     return dataset_dicts
 
 
-def create_test_scalar_dataset(output_url, num_rows, num_files=4, spark=None, partition_by=None):
+def create_test_scalar_dataset(output_url, num_rows, num_files=4, spark=None,
+                               partition_by=None, two_datetime_vals=False):
```
selitvin, line 163: Can you please clarify why did you choose to use alternating datatime values instead of a condition on an existing field? I think our test fixture data is not very well structured and hard to use (different fields are set using not clear nor intuitive logic).Can your test rely on one of the existing fields instead of adding more complexity to the data (unless you think your solution simplifies the testing code by alot)?

petastorm/tests/test_end_to_end.py
abditag2: Good point, fixed.

petastorm/tests/test_end_to_end.py
abditag2: done.

petastorm/tests/test_common.py
abditag2: modified to use an existing field.

bobingm:
I just noticed this change. Thank you for supporting this. Is that possible to support passing filters for all columns by setting `use_legacy_dataset` as `False` in `pyarrow.parquet.ParquetDataset`[1]?

[1] https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html?highlight=parquetdataset

PR #566, opened 2020-07-06
bake mnist data into docker image

abditag2:
The link to download mnist data is very slow and tests time out. this diff bakes the mnist data into the docker image.


abditag2:
> Hi @abditag2 - thanks for your contribution here! However, I would prefer that we do not bake data into the main container. If downloading MNIST data is an inconvenience, can we find an alternative here? How does storing it in the container fix this issue for you? If you need to keep it in a container, would you be willing to create a second `Dockerfile` that inherits from the main one and then use that?

@jsgoller1  The link to download the MNIST dataset is a university link and at times it is extremely slow i.e., 3kb/s slow. This diff bakes the data into the docker and for the tests, eliminates the data download requirements. However, I have updated the tests to use the download if the data is not present in the docker image. So, it does not overwrite previous behavior.

PR #567, opened 2020-07-08
Add imports to README example.

rb-determined-ai:
I found that the fully-functional hello world example code was great for approaching the project for the first time... but figuring out these imports was pretty time consuming, and I thought it might be nice to just embed them in the example.


rb-determined-ai:
@jsgoller1 cool!  I signed the CLA.  FWIW I am always in favor of copy/paste examples over Jupyter notebooks.

Also, I added another couple missing imports I found.

PR #568, opened 2020-07-09
Added additional kwargs for Spark Dataset Converter

tgaddair:
Adds support for:

- Shuffling
- Custom data loaders

### DIFF
--- a/petastorm/spark/spark_dataset_converter.py
+++ b/petastorm/spark/spark_dataset_converter.py
@@ -201,6 +201,8 @@ class SparkDatasetConverter(object):
             prefetch=None,
             num_epochs=None,
             workers_count=None,
+            shuffle_buffer_size=None,
+            make_reader_fn=None,
             **petastorm_reader_kwargs
     ):
         """Make a tensorflow dataset.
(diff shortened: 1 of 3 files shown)


petastorm/spark/spark_dataset_converter.py
```diff
@@ -200,6 +200,8 @@ def make_tf_dataset(
             prefetch=None,
             num_epochs=None,
             workers_count=None,
+            shuffle_buffer_size=None,
```
selitvin, line 203: Is `shuffle_buffer_size` effectively different from `shuffling_queue_capacity` in `make_torch_dataloader`? If not, we should probably call them the same.

tgaddair: Sounds good.  Do you want to go with `shuffling_queue_capacity` since it's already being used by `make_torch_dataloader`?

tgaddair:
@selitvin any idea what's causing these test failures?  They seem to be unrelated.

selitvin:
The failure is in the `test_tf_make_reader_fn`: `ValueError: directory url must be a string` - are you sure it's unrelated to your change.

selitvin:
Adding @mengxr for viz.

tgaddair:
@WeichenXu123, can you explain why support for `make_reader` was removed in https://github.com/uber/petastorm/commit/6c49ed5e6f6a193faf0701db46e886b9a5be5d5c#diff-b0f255caa777716ecea31adea665c124 ?

tgaddair:
@mengxr can you comment on the reason for 6c49ed5#diff-b0f255caa777716ecea31adea665c124?  We want to be able to use `make_reader` with Spark Dataset Converter.

mengxr:
https://github.com/uber/petastorm/pull/503#issue-386154352 

We removed `make_reader` because we want to support an input URL list. @WeichenXu123 's PR only added that support to `make_batch_reader`.

tgaddair:
@selitvin removed the `make_reader` support for now.  I think tests should be passing now, though there seems to be a failure due a timeout.  Is there an easy way to rerun the tests?

nightflight-dk:
greetings from MSFT, early perf tests of this branch show promising results, we look forward to seeing this in master

tgaddair:
Thanks for trying it out, @nightflight-dk.  This has now been merged into master.

PR #578, opened 2020-07-27
Add Python 3.6 to travis CI docker image

selitvin:


jsgoller1:
I don't know why we need to do this (not saying there isn't a reason, just that the description is blank so I have no context). LGTM though; will stamp once we discuss.

Issue #591, opened 2020-09-10
Schema inference does not apply filters to Metadata Discovery


dmcguire81:
Schema inference in `petastorm.reader.make_reader` utilizes `pyarrow.parquet.ParquetDataset` to discover metadata (via `petastorm.etl.dataset_metadata.get_schema_from_dataset_url`), but does not pass `filters` as an argument (see [the instantiation here](https://github.com/uber/petastorm/blob/master/petastorm/etl/dataset_metadata.py#L395)). This means that there is a bottleneck in both `make_reader` and `make_batch_reader` (since it uses the process of schema inference to try to give error messages) for any dataset with a large number of partitions, irrespective of the `filters` passed in to those functions. Essentially, adding that argument was not meaningful without using it for metadata, in addition to data.

dmcguire81:
Circling back to test, the `filters` argument does not appear to be passed to `pyarrow.parquet.ParquetManifest`, so can't actually speed up the discovery.

PR #592, opened 2020-09-10
Bugfix: multithreaded metadata deadlock

### LINKED ISSUE #590: Deadlock in multithreaded Parquet metadata discovery
### Summary

Interaction between Petastorm and S3FS seems to be unusable, and it's unclear what level of testing and exercise this has gotten within the Petastorm project, itself, and the wider community, because basic operations (`make_reader`, `make_batch_reader`) simply don't work at all. Breakdowns in the interaction could fall anywhere between this project, `pyarrow` (for `S3FSWrapper`) and `s3fs`, but we're consuming Petastorm, directly, so starting here.

### Tested Versions

This had to be tested on an earlier version of `s3fs` (`0.4.2`), because the more current versions (`>=0.5.0`) have a different problem with the `aiobotocore` wrapper leaking async coroutines into Petastorm. There will be a separate defect for that, and we'll take both to that project.

#### Setup
```sh
pip install petastorm==0.9.5
pip install s3fs==0.4.2
```

#### Test

```python
import pyarrow.parquet as pq
from petastorm.fs_utils import get_filesystem_and_path_or_paths, normalize_dir_url

dataset_url = 's3://<redacted>'

# Repeat basic steps that make_reader or make_batch_reader normally does
dataset_url = normalize_dir_url(dataset_url)
fs, path = get_filesystem_and_path_or_paths(dataset_url)

# Finished in seconds
dataset = pq.ParquetDataset(path, filesystem=fs, metadata_nthreads=1)
# Hung all night
dataset = pq.ParquetDataset(dataset_url, filesystem=fs, metadata_nthreads=10)
```

dmcguire81:
Fixes #590 and #591

### DIFF
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -382,7 +382,7 @@ def get_schema(dataset):
     return schema
 
 
-def get_schema_from_dataset_url(dataset_url_or_urls, hdfs_driver='libhdfs3'):
+def get_schema_from_dataset_url(dataset_url_or_urls, hdfs_driver='libhdfs3', filters=None):
     """Returns a :class:`petastorm.unischema.Unischema` object loaded from a dataset specified by a url.
 
     :param dataset_url_or_urls: a url to a parquet directory or a url list (with the same scheme) to parquet files.
@@ -392,7 +392,7 @@ def get_schema_from_dataset_url(dataset_url_or_urls, hdfs_driver='libhdfs3'):
     """
     fs, path_or_paths = get_filesystem_and_path_or_paths(dataset_url_or_urls, hdfs_driver)
 
-    dataset = pq.ParquetDataset(path_or_paths, filesystem=fs, validate_schema=False)
+    dataset = pq.ParquetDataset(path_or_paths, filesystem=fs, validate_schema=False, filters=filters)
 
     # Get a unischema stored in the dataset metadata.
     stored_schema = get_schema(dataset)
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -392,7 +392,7 @@ def get_schema_from_dataset_url(dataset_url_or_urls, hdfs_driver='libhdfs3'):
     """
     fs, path_or_paths = get_filesystem_and_path_or_paths(dataset_url_or_urls, hdfs_driver)
 
-    dataset = pq.ParquetDataset(path_or_paths, filesystem=fs, validate_schema=False, metadata_nthreads=10)
+    dataset = pq.ParquetDataset(path_or_paths, filesystem=fs, validate_schema=False)
 
     # Get a unischema stored in the dataset metadata.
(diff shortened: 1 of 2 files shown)


dmcguire81:
@selitvin how is this PR failing for code coverage in files that it did not change?

selitvin:
> @selitvin how is this PR failing for code coverage in files that it did not change?

Not sure. Some glitch that has to do probably with one of the source files have a line removed.

selitvin (commented):
Should we make the decision here automatically based on the filesystem used so we don't impact users of other file systems?

dmcguire81:
> Should we make the decision here automatically based on the filesystem used so we don't impact users of other file systems?

We could, but the `pyarrow.Filesystem` abstraction is basically broken, and it's up to you how much you want to have to maintain different code paths for different filesystems to work around that. Happy to change this or pursue a refactoring.

dmcguire81:
Would it look like a `defaultdict` with 1 entry for the broken filesystem, and a default of 10?

selitvin:
> but the pyarrow.Filesystem abstraction is basically broken

Can you clarify what are you referring to?

Another thought: would introducing serialization in `S3FSWrapper` guard underlying non-thread-safe code? That would be nicer than having to choose 1 vs 10 (which I agree is a maintenance headache)

dmcguire81:
> > but the pyarrow.Filesystem abstraction is basically broken
> 
> Can you clarify what are you referring to?

Sure, just that `pyarrow` assumes the `s3fs` code is threadsafe, when it clearly is not. That's they're problem to deal with, but doing the instantiation in `petastorm` makes it *also* your problem.

> Another thought: would introducing serialization in `S3FSWrapper` guard underlying non-thread-safe code? That would be nicer than having to choose 1 vs 10 (which I agree is a maintenance headache)

It could, and I think fixing the problem within `pyarrow` is even better than fixing it with `petastorm`, but probably not as good as [fixing it with s3fs](https://github.com/dask/s3fs/issues/365). Unfortunately, I'm having trouble navigating the defect reporting process with Apache Arrow, since it's based on Jira, not, GitHub. It's unclear if I'll need an additional login (and associated legal approval on my side), so if you can navigate that process and see your way through it, that would be *great*!

dmcguire81:
@selitvin the [conversation with s3fs](https://github.com/dask/s3fs/issues/365#issuecomment-691221192) isn't getting a lot of traction, so I'm wary of trying to thread piecemeal fixes through `petastorm`, instead of just isolating the problem by instantiating the `pyarrow.FileSystem` as a client. I'm going to close this out and open a PR to pass in the FS as an argument, instead.

dmcguire81:
Turns out trying to make metadata discovery more efficient by adding `filters` is what caused the defect to crop up, in the first place, as the [deadlock doesn't exist without the `filters` argument](https://github.com/dask/s3fs/issues/365#issuecomment-691242646). @selitvin there was no way that `petastorm` could have found or anticipated this defect.

Issue #593, opened 2020-09-10
The boto config used by s3fs is not parameterizeable when instantiated by Petastorm


dmcguire81:
The problem is that their stock configuration uses the [legacy retry mode](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html#available-retry-modes), which is not the recommendation from AWS. Since Petastorm instantiates the `s3fs` client when using either `make_reader` or `make_batch_reader`, there's no opportunity to configure it correctly, and brute-force changes to the `~/.aws/config` file impacts *any* code using `boto3`, even if unrelated, so there is no fine-grained control to the number of retries by use case.

selitvin:
Got it. I assume our idea of parametrizing make_read/make_batch_reader with a file-system instance would solve this. RIght?

dmcguire81:
The merged code solved this, but parameterizing with a `FileSystem` might allow me to work-around the deadlock in the interaction between `pyarrow` and `s3fs`. As mentioned, I'll pull you into that ongoing conversation.

PR #594, opened 2020-09-10
Parameterize factory methods with s3 configs

dmcguire81:
Implements #593


selitvin (approved):
Please add a line to `docs/release-notes.rst`. Otherwise, I think this is good. Let me know if we you would like me to land.

dmcguire81:
Updated. This can be merged.

dmcguire81:
@selitvin how do I make the erroneous Codecov errors go away?

selitvin:
I think it's about codecov configuration file. Did not have time to fix this - I am just looking at the actual numbers and if they make sense, then force land it.

Issue #595, opened 2020-09-10
Expose the flag to disable Ømq copy buffers


dmcguire81:
One of our engineers found an optimization involving disabling ZeroMQ copy buffers in the `ProcessWorker`, but this is not exposed in the top-level factory methods, `make_reader` and `make_batch_reader`. It's useful, and probably should be.

selitvin:
Yep. Agreed. Can you please elaborate what is the issue?

dmcguire81:
There's nothing new to report: better throughput, in exchange for less predictable memory consumption, is a good trade-off for us, and probably for others, so it should be available in the factory methods.

dmcguire81:
While we're talking, though, it seems like the factory methods only work seamlessly if on whatever storage Uber happens to use (guessing HDFS, given `libhdfs` is the only config given consistently elevated visibility). The experience on S3 is pretty bad.

dmcguire81:
We're considering dropping down to the `petastorm.reader.Reader` interface to deal with having to patch `s3fs` and `pyarrow` for the respective shortcomings (or this code base to work around them), but wanted to get your take on a refactoring to the factory methods, themselves, since it seems like a losing battle to offer all the flexibility users will need *and* still all do the instantiation yourself.

selitvin:
Your diagnosis is correct. We did not use petastorm with s3 internally, hence this is not a polished feature. I can help if you can point me to the issues you run into. I did verify basic functionality but not more then that.

selitvin:
I saw that you created #594. If this is not sufficient (curious what else would we need), perhaps we can have `make_reader/make_batch_reader` take an instance of already constructed filesystem object?

dmcguire81:
We built up a bit of a backlog while waiting on legal to approve submitting PRs back, but they're all in there, now, so you see everything that's been needed, so far (except for `s3fs==0.5.0` throwing `TypeError: 'coroutine' object is not iterable` from `make_batch_reader`, which I just found today, by accident, trying to build you a repro case.

I think passing the `pyarrow.Filesystem` is a great option, since the rest of the code that's hidden (e.g., instances of `petastorm.worker_pool.*`) is under your direct control, so won't be as much of a maintenance burden to keep "under the hood".

dmcguire81:
#590 was actually the real killer, but that addresses every problem we've seen, so far, *except* this one, so we could get it all done in 2 PRs, probably.

dmcguire81:
[...]

PR #596, opened 2020-09-10
Add a flag to factory methods to allow zmq copy buffers to be disabled

dmcguire81:
Addresses #595

### DIFF
--- a/petastorm/workers_pool/process_pool.py
+++ b/petastorm/workers_pool/process_pool.py
@@ -311,6 +311,7 @@ class ProcessPool(object):
             'items_consumed': self._ventilated_items,
             'items_produced': self._ventilated_items_processed,
             'items_inprocess': self._ventilated_items - self._ventilated_items_processed,
+            'zmq_copy_buffers': self._zmq_copy_buffers
         }
 
 
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -7,6 +7,7 @@ Release notes
 Release 0.9.6 (unreleased)
 ==========================
 - `PR 588 <https://github.com/uber/petastorm/pull/588>`_: ``make_reader`` can now open a parquet dataset from a subdirectory in an s3 bucket.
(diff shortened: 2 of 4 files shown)


selitvin (commented):
Let's add a test that would at least exercise zmq_copy_buffers=False.
Also, please add a line to `docs/release-notes.rst`

dmcguire81:
I still owe you a test, here, so I'll get to this, today.

petastorm/workers_pool/process_pool.py
```diff
@@ -311,6 +311,7 @@ def diagnostics(self):
             'items_consumed': self._ventilated_items,
             'items_produced': self._ventilated_items_processed,
             'items_inprocess': self._ventilated_items - self._ventilated_items_processed,
+            'zmq_copy_buffers': self._zmq_copy_buffers
```
dmcguire81, line 314: I extended the Worker Pool `diagnostics()` rather than read into a chain of private fields in the test.

dmcguire81:
@selitvin let me know if this test works for you

selitvin:
@dmcguire81: looks good. Let's fix whatever flake8 complaints about and land it.

dmcguire81:
@selitvin can you restart the Travis CI build? It looks like a brittle test is timing out, which seems unrelated.

selitvin:
> @selitvin can you restart the Travis CI build? It looks like a brittle test is timing out, which seems unrelated.

Hmmm. This is annoying. Did not see these timeouts before... Restarted.

selitvin:
These test timeouts appear to be in the similar place as the hanging you observed with s3. Do you think they are related?

dmcguire81:
> These test timeouts appear to be in the similar place as the hanging you observed with s3. Do you think they are related?

Could be. I'll pull you into the ongoing discussion with `pyarrow` and `s3fs` maintainers, since they asked for a reproducer that you may be able to build more quickly than me, if you're more familiar. Needing to work around this deadlock is also related to our discussion of parameterizing the `FileSystem` instance.

dmcguire81:
@selitvin https://github.com/dask/s3fs/issues/365

selitvin (approved):

PR #598, opened 2020-09-16
Use path with bucket name if it's an s3 path and a custom filesystem.

acmore:
When using s3 compatible object storage, we would like to pass in a S3FileSystem overriding the endpoint. The path should include the bucket name when it's s3 object storage.


selitvin (commented):
Can you please add a line describing the fix to the release notes (docs/release-notes.rst)?

acmore:
> Can you please add a line describing the fix to the release notes (docs/release-notes.rst)?

Sure. Added.

Issue #597, opened 2020-09-14
can we use s3 path here instead of hdfs?


p9anand:
from petastorm.spark import SparkDatasetConverter, make_spark_converter

# specify a cache dir first.
# the dir is used to save materialized spark dataframe files
**_spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF, 'hdfs:/...')_**

**### _can we use s3 path here instead of hdfs?_**

df_train, df_test = ... # `df_train` and `df_test` are spark dataframes
model = Net()

# create a converter_train from `df_train`
# it will materialize `df_train` to cache dir. (the same for df_test)
converter_train = make_spark_converter(df_train)
converter_test = make_spark_converter(df_test)

# make a pytorch dataloader from `converter_train`
with converter_train.make_torch_dataloader() as dataloader_train:
    # the `dataloader_train` is `torch.utils.data.DataLoader` object
    # we can train model using the `dataloader_train`
    train(model, dataloader_train, ...)
    # when exiting the context, the reader of the dataset will be closed

# the same for `converter_test`
with converter_test.make_torch_dataloader() as dataloader_test:
    test(model, dataloader_test, ...)

# delete the cached files of the dataframes.
converter_train.delete()
converter_test.delete()

selitvin:
Did you try doing that? What did you observe?

p9anand:
yes. i tried following codes:

from petastorm.spark import SparkDatasetConverter, make_spark_converter

# specify a cache dir first.
# the dir is used to save materialized spark dataframe files
spark.conf.set(SparkDatasetConverter.PARENT_CACHE_DIR_URL_CONF, 's3a://****/*****/')

df_train, df_test = train_df, test_df # `df_train` and `df_test` are spark dataframes
model = Net()

# create a converter_train from `df_train`
# it will materialize `df_train` to cache dir. (the same for df_test)
converter_train = make_spark_converter(df_train)
converter_test = make_spark_converter(df_test)

# make a pytorch dataloader from `converter_train`
with converter_train.make_torch_dataloader() as dataloader_train:
    # the `dataloader_train` is `torch.utils.data.DataLoader` object
    # we can train model using the `dataloader_train`
    train(model, dataloader_train, ...)
    # when exiting the context, the reader of the dataset will be closed

# the same for `converter_test`
with converter_test.make_torch_dataloader() as dataloader_test:
    test(model, dataloader_test, ...)

# delete the cached files of the dataframes.
[...]

PR #601, opened 2020-09-17
Allow users to use s3, s3a and s3n protocols when saving / reading datasets

selitvin:
s3, s3a and s3n url protocols can be explicitly specified when saving petatorm datasets.

Fixed a bug on petastorm dataset write execution path previously preventing writing directly to s3 buckets.

Tested: modified `examples/generate_external_dataset.py` and `examples/python_hello_world.py` to write/read from s3 bucket using s3a and s3n buckets (wasn't able to properly configure s3 authentication to check that). Was able to write/read data successfully.

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -9,6 +9,7 @@ Release 0.9.6 (unreleased)
 - `PR 588 <https://github.com/uber/petastorm/pull/588>`_: ``make_reader`` can now open a parquet dataset from a subdirectory in an s3 bucket.
 - `PR 594 <https://github.com/uber/petastorm/pull/594>`_: Parameterize factory methods with s3 configs
 - `PR 596 <https://github.com/uber/petastorm/pull/596>`_: Add a flag to factory methods to allow zmq copy buffers to be disabled
+- `PR 601 <https://github.com/uber/petastorm/pull/601>`_: Allow user to use s3, s3a and s3n url schemes when writing datasets.
 
 Release 0.9.5
 ==========================
--- a/petastorm/etl/dataset_metadata.py
+++ b/petastorm/etl/dataset_metadata.py
@@ -18,17 +18,17 @@ import os
 from concurrent import futures
 from contextlib import contextmanager
 from operator import attrgetter
+from urllib.parse import urlparse
 
+from packaging import version
 from pyarrow import parquet as pq
 from six.moves import cPickle as pickle
-from six.moves.urllib.parse import urlparse
 
 from petastorm import utils
 from petastorm.compat import compat_get_metadata, compat_make_parquet_piece
 from petastorm.etl.legacy import depickle_legacy_package_name_compatible
-from petastorm.fs_utils import FilesystemResolver, get_filesystem_and_path_or_paths
+from petastorm.fs_utils import FilesystemResolver, get_filesystem_and_path_or_paths, get_dataset_path
 from petastorm.unischema import Unischema
-from packaging import version
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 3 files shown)


selitvin:
@dmcguire81 , can you please help me with this small review as you have a lot of experience with s3 and petastorm. Much appreciated!

dmcguire81:
@selitvin this will work, but I'm not sure you want to say that `petastorm` "supports" those protocols. Please see the comment in #597. When the author asks "instead of hdfs", they specifically think that `s3` is some really, really old, Hadoop-specific notion of S3 storage (in 10 years in Big Data, I've not run across anyone using it), and it's not. It would be trivial, and conceptually less confusing, for them to substitute the protocol in their own URLs, as if to say "`s3a` in `hadoop-aws` is just `s3` in `boto3`".

Now, if you want to do that for them, it's certainly *safe* to do so, because `boto3`'s treatment of the `s3` protocol should be strictly more powerful than `s3n` (no file size limitation), and equivalent to `s3a`. However, introducing this meaningless distinction in Python may just make *more* users coming from Hadoop/AWS assume that `s3` is the older, incompatible protocol, I'm afraid.

dmcguire81:
If you still want to move forward after the Issue author responds, I have some `s3a` prefixes I can also test against, but `s3n` is going to be a challenge.

selitvin:
Thank you for the comment! Not sure I understand. Is your comment about "petastorm supports s3/s3n/s3a" refers to the documentation, or code? 

According to your suggestion would user always specify `s3://..." and petastorm should replace it `s3a://` when passing it down to `pyarrow.filesystem.S3FSWrapper` (because that's how boto3 is doing it?), or I misunderstood you?

dmcguire81:
@selitvin `s3a` and `s3n` are meaningless protocol specifiers in the Python stack. The only thing `boto3` understands is `s3`. However, how it *treats* that protocol is exactly what the author has in mind for how Hadoop treats `s3a` (it's AWS native SDK code, and there is no arbitrary file limit). In short, asking the author to use `s3` in place of `s3a` in all their URLs, because it's 100% compatible with the storage written, is reasonable, in my view.

dmcguire81:
> Thank you for the comment! Not sure I understand. Is your comment about "petastorm supports s3/s3n/s3a" refers to the documentation, or code?

Both, since there is implied support in the code.

> According to your suggestion would user always specify `s3://..." and petastorm should replace it `s3a://`when passing it down to`pyarrow.filesystem.S3FSWrapper` (because that's how boto3 is doing it?), or I misunderstood you?

Petastorm should stay the same, in this scenario, except for maybe giving a hint that those protocols don't mean anything in Python.

dmcguire81 (approved):
Looks safe. Probably need to give some thought to how to communicate to users coming from Hadoop.

selitvin:
Updated release notes to the best of my understanding of the problem. Please let me know if it does not make sense and I'll fix it.

dmcguire81 (approved):
Looks good. What's your general release cadence?

selitvin:
> Looks good. What's your general release cadence?

On demand. I think we have enough goodies to cut a release. Unless you have more stuff coming up, I can do it until tomorrow EOD.

dmcguire81:
> > Looks good. What's your general release cadence?
> 
> On demand. I think we have enough goodies to cut a release. Unless you have more stuff coming up, I can do it until tomorrow EOD.

Nothing else pending (PyArrow changes look to be extensive). Tomorrow works.

selitvin:
@dmcguire81 : FYI: 0.9.6rc0 is available on pypi.

PR #606, opened 2020-09-24
Fix docker file

selitvin:
Adding make and skbuild to the image


jsgoller1 (approved):
filthy_frank_clapping

PR #615, opened 2020-10-21
Raising an error if an empty list of columns is selected by a user

selitvin:
Makes it easier to detect user logic error, as it makes no sense to create a reader that loads no data from storage.

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -421,6 +421,8 @@ class Reader(object):
             fields = schema_fields if isinstance(schema_fields, collections.Iterable) else None
 
         storage_schema = stored_schema.create_schema_view(fields) if fields else stored_schema
+        if len(storage_schema.fields) == 0:
+            raise RuntimeError(f"No fields matching the criteria '{fields}' were found in the dataset {dataset_path}.")
         if transform_spec:
             self.schema = transform_schema(storage_schema, transform_spec)
         else:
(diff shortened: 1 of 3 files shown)


petastorm/tests/test_parquet_reader.py
```diff
@@ -165,8 +162,7 @@ def test_invalid_and_valid_column_names(scalar_dataset, reader_factory):
     with reader_factory(scalar_dataset.url, schema_fields=requested_fields) as reader:
         sample = next(reader)._asdict()
         assert len(sample) == 1
-        with pytest.raises(KeyError):
-            assert sample[bad_field] == ""
+        assert bad_field not in sample
```
selitvin, line 165: Piggybagging this small cleanup. Not directly related to the PR

jsgoller1 (approved):
LGTM; sorry for the delay.

petastorm/reader.py
```diff
@@ -421,6 +421,8 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
             fields = schema_fields if isinstance(schema_fields, collections.Iterable) else None
 
         storage_schema = stored_schema.create_schema_view(fields) if fields else stored_schema
+        if len(storage_schema.fields) == 0:
```
jsgoller1, line 424: Nit: `if not storage_schema.fields` should work here since the `.fields` attribute is an `OrderedDict`.

selitvin: That's true. I don't mind either, but frankly prefer the explicit check since otherwise it may be interpreted as checking if the attribute is None.

PR #617, opened 2020-11-07
Deprecating `pyarrow_serialize` argument of `petastorm.make_reader`.

selitvin:
pyarrow serialization is no longer suppoted. It actually has no merit as python3's pickle is faster then pyarrow serialization with numpy arrays.


petastorm/reader.py
```diff
@@ -152,9 +150,9 @@ def make_reader(dataset_url,
         reader_pool = ThreadPool(workers_count, results_queue_size)
     elif reader_pool_type == 'process':
         if pyarrow_serialize:
-            serializer = PyArrowSerializer()
-        else:
-            serializer = PickleSerializer()
+            warnings.warn("pyarrow_serializer was deprecated and will be removed in future versions. "
+                          "The argument no longer has any effect.")
```
Ivan-Dimitrov, line 154: Isn't it more accurate to say `was deprecated and removed. The argument no longer has any effect.`.

S

Ivan-Dimitrov (approved):
great deleting

petastorm/reader.py
selitvin: Strictly speaking we didn't remove it from the API. It does not have any effect anymore and that's mentioned in the message. If we would remove the argument, the existing code would break.

PR #618, opened 2020-11-07
Deprecate compat library since we no longer support pre pyarrow 0.17

selitvin:
This PR removes compat.py and cleans up the code. The PR should be a noop with pyarrow >= 0.15


Ivan-Dimitrov (approved):
lgtm

PR #628, opened 2020-12-06
Fixed docstring in NGram class: timestamp_field argument illustration…

ritwikbera:
The usage examples shown in the docstring of the NGram class have an error. Entering the timestamp parameter as a string (i.e. 'id') doesn't work. It should be a Unischema Field. (i.e. entered as TestSchema.id). This pull request reflects the necessary change in the docstring.


selitvin (approved):
Thanks!

PR #630, opened 2020-12-21
Add unit tests for compress in random shuffling buffer

chongxiaoc:
~Compress remaining shuffling buffer should use remained size, that is, `self._size`.~

`self.size` actually is a property decorator defined afterwards.
I just change it to keep consistent with other places of code to improve readability.

Also, I added some unit tests to check compress results.

### DIFF
--- a/petastorm/reader_impl/pytorch_shuffling_buffer.py
+++ b/petastorm/reader_impl/pytorch_shuffling_buffer.py
@@ -260,7 +260,7 @@ class BatchedRandomShufflingBuffer(BatchedShufflingBufferBase):
             # Before we can append a new batch, we compress the remaining samples
             for k, v in enumerate(self._items):
                 # We need to clone the right-side to avoid racing conditions
-                self._items[k][:self.size] = self._items[k][self._random_indices[self.next_sample_head:]].clone()
+                self._items[k][:self._size] = self._items[k][self._random_indices[self.next_sample_head:]].clone()
         self._random_indices = None
         self.next_sample_head = 0
 
--- a/petastorm/tests/test_shuffling_buffer.py
+++ b/petastorm/tests/test_shuffling_buffer.py
@@ -234,3 +234,41 @@ def test_longer_random_sequence_of_queue_ops(buffer_type):
             # Make sure never get to less than `min_after_retrieve` elements
             assert 80 <= q.size
             _retrieve(q)
+
+
+@pytest.mark.parametrize('buffer_type', RANDOM_SHUFFLING_BUFFERS)
+def test_random_shuffling_buffer_compress(buffer_type):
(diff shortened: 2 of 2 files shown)


chongxiaoc:
@selitvin I cannot request reviewer on the right side panel, can you help me review or assign it? Thanks.

selitvin:
Thanks for the fix. Really surprised we didn't catch that with a unit-test. Can you see if a test case is missing? If not, please let me know I'll take a stab at it.

chongxiaoc:
> Thanks for the fix. Really surprised we didn't catch that with a unit-test. Can you see if a test case is missing? If not, please let me know I'll take a stab at it.

@selitvin 
It seems that in `test_shuffling_buffer.py`, only result length,  can_add, can_retrieve are tested. There is no test to verify the output of `add_many` with `remaining shuffle buffer` should be valid. Maybe I missed some points, I was new to petastorm. :)

selitvin:
Appears to be a miss then! Would you be able to extend the test to cover this line with the typo or you'd need a hand with that?

chongxiaoc:
> Appears to be a miss then! Would you be able to extend the test to cover this line with the typo or you'd need a hand with that?

Sure. I will start with that and reach out to you if need help.

selitvin:
Appreciate it! Would be glad to help, but I think it should be pretty straight forward. Don't know how did we miss that...

chongxiaoc:
@selitvin 
Actually this is not a typo in original code.
`self.size` is defined as a property decorator afterwards.
But I just recommend to change it to be consistent with all other places in the code using `self._size`

Please feel free to drop the first commit.

I also added necessary unit tests to check compress results in second commit, please take a look and let me know anything to change.

chongxiaoc:
@selitvin How to re-trigger the test? I don't quite understand why build can fail just for a single character change.

selitvin:
Appears to be a linting issue:
```
petastorm/tests/test_shuffling_buffer.py:262: [C0325(superfluous-parens), ] Unnecessary parens after 'not' keyword

```
https://travis-ci.com/github/uber/petastorm/jobs/464041086

You should be able to repro locally:
```bash
pylint --rcfile=.pylintrc petastorm examples -f parseable -r n
```

chongxiaoc:
> Appears to be a linting issue:
> 
> ```
> petastorm/tests/test_shuffling_buffer.py:262: [C0325(superfluous-parens), ] Unnecessary parens after 'not' keyword
> ```
> 
> https://travis-ci.com/github/uber/petastorm/jobs/464041086
> 
> You should be able to repro locally:
> 
> ```shell
> pylint --rcfile=.pylintrc petastorm examples -f parseable -r n
> ```

Got it!

chongxiaoc:
@selitvin Fixed.

Issue #633, opened 2021-01-15
Unischema variable shape error


berendjansen:
Hi, 

I'm trying to use the spark dataset converter to build a tensorflow dataset. I am training an LSTM on sequences with variable length. 

In the transform spec function I pass to `make_tf_dataset`, I set `shape=(None, 34)` for the UnischemaField:
```
transform_spec_fn = TransformSpec(
  transform_row, 
  edit_fields=[('numpy_sequence', np.float32, shape=(None, 34), True)], 
  selected_fields=['numpy_sequence']
)
```

This is to indicate that sequences can be of variable length, which corresponds to the documentation of `UnischemaField`:

```
shape: shape of the multidimensional array. None value is used to define a dimension with variable number of
elements. E.g. ``(None, 3)`` defines a point cloud with three coordinates but unknown number of points.
```

However, when I'm trying to build the dataset, I get an error saying that my field does not have the shape `(None, 34)`. 
```
with converter_train.make_tf_dataset(transform_spec=transform_spec_fn,) as train_dataset:
    ...
```

Results in:
```
Traceback (most recent call last):
  File "/databricks/python/lib/python3.7/site-packages/petastorm/workers_pool/thread_pool.py", line 62, in run
    self._worker_impl.process(*args, **kargs)
  File "/databricks/python/lib/python3.7/site-packages/petastorm/arrow_reader_worker.py", line 167, in process
    lambda: self._load_rows(parquet_file, piece, shuffle_row_drop_partition))
  File "/databricks/python/lib/python3.7/site-packages/petastorm/cache.py", line 39, in get
    return fill_cache_func()
  File "/databricks/python/lib/python3.7/site-packages/petastorm/arrow_reader_worker.py", line 167, in <lambda>
    lambda: self._load_rows(parquet_file, piece, shuffle_row_drop_partition))
  File "/databricks/python/lib/python3.7/site-packages/petastorm/arrow_reader_worker.py", line 219, in _load_rows
    .map(lambda x, f=field: self._check_shape_and_ravel(x, f))
  File "/databricks/python/lib/python3.7/site-packages/pandas/core/series.py", line 3630, in map
    new_values = super()._map_values(arg, na_action=na_action)
  File "/databricks/python/lib/python3.7/site-packages/pandas/core/base.py", line 1154, in _map_values
    new_values = map_f(values, mapper)
  File "pandas/_libs/lib.pyx", line 2327, in pandas._libs.lib.map_infer
[...]

PR #634, opened 2021-01-15
Raise an explicit error when TransformSpec is given a shape with a variable dimension

selitvin:
Records returned by `make_batch_reader` are returned as a dictionary of coalesced numpy arrays. In order to properly batch multiple rows together, all fields must have the same dimension. If a user passes an shape that have some variable dimensions, multiple rows won't coalesce properly.

This PR issues a clear message.

See also: #633

### DIFF
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -177,7 +177,7 @@ class ArrowReaderWorker(WorkerBase):
             raise ValueError(f'All dimensions of a shape: {field.shape} in: {field.name} field must be constant. '
                              f'If a dimension is variable, we won\'t be able to coalesce rows when preparing a batch.')
 
-        if x.shape != field.shape:
+        if x.shape != tuple(field.shape):
             raise ValueError('field {name} must be the shape {shape}'
                              .format(name=field.name, shape=field.shape))
         if not x.flags.c_contiguous:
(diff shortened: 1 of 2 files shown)


petastorm/arrow_reader_worker.py
```diff
@@ -173,6 +173,10 @@ def process(self, piece_index, worker_predicate, shuffle_row_drop_partition):
     def _check_shape_and_ravel(x, field):
         if not isinstance(x, np.ndarray):
             raise ValueError('field {name} must be numpy array type.'.format(name=field.name))
+        if None in field.shape:
```
WeichenXu123, line 176: Should we also check dim value -1 in shape ?

petastorm/arrow_reader_worker.py
```diff
@@ -173,6 +173,10 @@ def process(self, piece_index, worker_predicate, shuffle_row_drop_partition):
...
+            raise ValueError(f'All dimensions of a shape: {field.shape} in: {field.name} field must be constant. '
+                             f'If a dimension is variable, we won\'t be able to coalesce rows when preparing a batch.')
+
         if x.shape != field.shape:
```
WeichenXu123, line 180: Should we use `x.shape != tuple(field.shape)`, in case of field.shape may be a list ?

petastorm/arrow_reader_worker.py
selitvin: Not sure. `None` is a generally, valid element in UnischemaField's shape field, but -1 is not. If some validation on invalid range on dimensions should be done, I would think that it should be done sometime during the construction of UnischemaField?

WeichenXu123 (approved):
LGTM

Issue #638, opened 2021-01-21
Implementing Asynchronous Dataloader


chongxiaoc:
Existing dataloader implementation is sequential, for common use case, it limits training speed if data shuffle part is slow. This happens when user defines a pretty large `shuffling queue capacity`, whenever dataloader reads next row group, `shuffling queue buffer` has to compress and reshuffle.

An idea of using asynchronous dataloader is to move parquet file reading, shuffling and producing batches into an asynchronous thread, maintain a batch queue for consumer (training thread) and producer (dataloader thread). In this case, during main thread training and backpropogragion stages, asynchronous can keep generating batches, to overlap the cost.

We have an idea of introducing asynchronous datalaoder based on the prototype for existing work. By looking at current code base, there are two ways to introduce it:
1. Add an option `is_async` into `LoaderBase` class, also other necessary variables for asynchronous operations.
 https://github.com/uber/petastorm/blob/cf1159dc04416ed737eec25bcecef5d5aafa805a/petastorm/pytorch.py#L132
This would introduce all async related logics into existing dataloader, for both non-batch and batch versions, which would make the dataloader implementation complicated.

2. Create a new `asyncLoaderBase`, also implement `AsyncDataLoader ` and `AsyncBatchedDataLoader` classes. In this case user has to create select dataloader explicitly for sync or async usage.

@selitvin What do you think about above idea? And what your suggestions to go in future to merge this work into petastorm?

FYI @tgaddair

chongxiaoc:
Had a discussion offline, start to work on moving the bottleneck `data shuffling part` into an asynchronous thread.
https://github.com/uber/petastorm/blob/cf1159dc04416ed737eec25bcecef5d5aafa805a/petastorm/pytorch.py#L373

And see how it can hide/overlap the latency.

PR #641, opened 2021-01-25
Enable async shuffle for BatchedDataLoader

chongxiaoc:
Add an asynchronous thread to shuffle batches in background. Main thread consumes ready shuffled batches in the queue. 
Related to #638


petastorm/pytorch.py
```diff
@@ -420,5 +458,17 @@ def __enter__(self):
...
+            # Close async shuffle thread
+            self.close_async_thread()
         self.reader.stop()
         self.reader.join()
+
+    def close_async_thread(self):
+        self.finished_event.set()
+        try:
+            # Free buffer to allow worker to retry
+            self.q.get_nowait()
```
tgaddair, line 471: Not sure why this is needed.  Is this because the producer may be blocked trying to add an element to the queue?  Would it make sense to just clear the queue instead?

petastorm/pytorch.py
```diff
@@ -304,6 +311,13 @@ def __init__(self, reader, batch_size=1,
         self._batch_acc = []
         self.shuffling_queue_capacity = shuffling_queue_capacity
         self._in_iter = None
+        self.q = None
+        if async_batch_queue_sz > 0:
+            self.q = Queue(async_batch_queue_sz)
+            self.finished_event = Event()
```
tgaddair, line 317: Should make these member variables "private" by adding an `_` to the front.

petastorm/pytorch.py
chongxiaoc: It is for the boundary case, len(q) == q_size, and producer is waiting on q.put(), cannot finish join() if consumer doesn't consume something in the queue.

petastorm/pytorch.py
```diff
@@ -420,5 +466,17 @@ def __enter__(self):
...
         self.reader.join()
+
+    def close_async_thread(self):
+        self.finished_event.set()
+        try:
+            # Free buffer to allow worker to retry
+            self.q.get_nowait()
+        except Empty:
+            pass
+        self.shuffle_thread.join()
```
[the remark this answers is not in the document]
chongxiaoc: I removed the `self.started`, it seems unnecessary.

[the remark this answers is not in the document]
chongxiaoc: Wait a sec, you are right for the `loader resue` part, then need to reset this flag.

chongxiaoc:
@tgaddair I fixed above comments, most of changes are related to make async shuffle thread work for reusing dataloader for the same reader multiple times. In that case, we have to launch a new async thread for every new pass of files (async thread has exited for previous pass when reach end of data).

Please take a look, thanks.

chongxiaoc:
@selitvin Any comments to debug the dockercloud-ci failure? Link is not working to me, showing 404.

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
...
+                if b is None:
+                    # None is last data in async batch queue, which means that async shuffle thread
+                    # has reached the end of data.
+                    self._started = False
+                    break
+                yield b
+        else:
+            # Sequentially shuffle data and produce batches in frontend.
+            for b in self._iter_impl_worker():
+                yield b
```
selitvin, line 397: nit: we don't need to support py2 anymore , so would a slightly nicer:
```python
yield from self._iter_impl_worker
```
work?

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
...
+            if self._finished_event.is_set():
+                break
+            self._async_batch_queue.put(b)
+        else:
+            self._async_batch_queue.put(None)
+
     def _iter_impl(self):
+        if self._async_batch_queue:
+            # Get shuffled batches from async shuffle thread
+            if not self._started:
```
selitvin, line 379: If you think it's worth it to start the thread early (in the constructor), can we factor out this code so we don't have duplication?

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
...
+                self._shuffle_thread = Thread(target=self._worker)
+                self._shuffle_thread.daemon = True
+                self._started = True
+                self._shuffle_thread.start()
+            while True:
+                b = self._async_batch_queue.get()
+                if b is None:
+                    # None is last data in async batch queue, which means that async shuffle thread
+                    # has reached the end of data.
+                    self._started = False
```
selitvin, line 391: Should we join here (and maybe self._shuffle_thread=None) for the sake of leaving a clean state on exit? Should we actually use `_close_async_thread` here?

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
 
         return shuffling_buffer, other_shuffling_buffer, instantiate_buffer_fn
 
+    def _worker(self):
+        for b in self._iter_impl_worker():
```
selitvin, line 369: For readability, consider `b` -> `batch` here and elsewhere.

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
...
+            self._async_batch_queue.put(None)
+
     def _iter_impl(self):
+        if self._async_batch_queue:
+            # Get shuffled batches from async shuffle thread
+            if not self._started:
+                # If reuse dataloader in a new pass, async shuffle thread has to be relaunched.
+                # It is finished in previous pass when reaching end of data.
+                self._shuffle_thread = Thread(target=self._worker)
+                self._shuffle_thread.daemon = True
```
selitvin, line 383: Is there a reason for the thread being daemonic? My concern is that random exceptions might show up on shutdown if we don't properly shutdown threads.

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -230,12 +240,15 @@ def test_mem_cache_num_epochs_without_mem_cache_error(two_columns_non_petastorm_
 @pytest.mark.parametrize('shuffling_queue_capacity', [20, 0])
 @pytest.mark.parametrize('reader_factory', [make_batch_reader, make_reader])
 @pytest.mark.parametrize('num_epochs', [1, 2, 3, None])
+@pytest.mark.parametrize('async_batch_queue_capacity', [0, 2, 4])
```
selitvin, line 243: The test is launched 2x2x4x3=48 times. Let's consolidate the parametrization to actually interesting cases and reduce test run time:
```python
@pytest.mark.parametrize('shuffling_queue_capacity,reader_factory,num_epochs,async_batch_queue_capacity', [[set1], [set2],...]
```

petastorm/pytorch.py
```diff
@@ -348,7 +365,38 @@ def _instantiate_noop_buffer():
...
+            self._async_batch_queue.put(b)
+        else:
+            self._async_batch_queue.put(None)
+
     def _iter_impl(self):
+        if self._async_batch_queue:
+            # Get shuffled batches from async shuffle thread
+            if not self._started:
+                # If reuse dataloader in a new pass, async shuffle thread has to be relaunched.
+                # It is finished in previous pass when reaching end of data.
```
selitvin, line 381: Is the restart scenario covered by a unit test?

selitvin (commented):
1. Do you think we can factor out new functionality into a separate class (separation of concerns: the queue thread / rest of the data loader) to keep class size more manageable?
2. Test cases become hairy with the parallelism introduced to the DataLoader implementation with some potential for race conditions. Would like to make sure we cover cases like cases like : 
 - We exit DataLoader context while shuffling queue is full
 - We exit DataLoader without fetching any samples
 - An error is raised somewhere within worker thread
 - possibly more cases...

petastorm/pytorch.py
```diff
@@ -289,6 +292,10 @@ def __init__(self, reader, batch_size=1,
         :param transform_fn: an optional callable to convert batches from the reader to PyTorch tensors
         :param shuffling_queue_capacity: Queue capacity is passed to the underlying :class:`tf.RandomShuffleQueue`
           instance. If set to 0, no shuffling will be done.
+        :param async_batch_queue_capacity: Number of batches to be produced in advance. If async_batch_queue_capacity
```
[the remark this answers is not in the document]
chongxiaoc: I don't mind for either case. Since I'm not sure the common data size of using BatchedDataLoader.  @tgaddair what do you think.

petastorm/pytorch.py
chongxiaoc: I see. Make sense to me.

petastorm/pytorch.py
chongxiaoc: This is covered by using multiple epochs in the unit test, for example:
```
for _ in range(num_epochs):
            # Check if dataloder works correctly with multiple passes.
            _check_simple_reader(loader, synthetic_dataset.data, TORCH_BATCHABLE_FIELDS - {TestSchema.sensor_name})
```

The 2nd and 3rd epoch will hit the restarting case, because previous round of data read already completes.

petastorm/pytorch.py
chongxiaoc: I see, I will think about it.

petastorm/pytorch.py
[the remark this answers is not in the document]
tgaddair: I think the only question for enabling it by default is: what should the default capacity be?  If we can think of a good default, then I think that's reasonable.

chongxiaoc:
> 1. Do you think we can factor out new functionality into a separate class (separation of concerns: the queue thread / rest of the data loader) to keep class size more manageable?

Good point, I will think about it.

> 2. Test cases become hairy with the parallelism introduced to the DataLoader implementation with some potential for race conditions. Would like to make sure we cover cases like cases like :
> 
> * We exit DataLoader context while shuffling queue is full
> * We exit DataLoader without fetching any samples
> * An error is raised somewhere within worker thread
> * possibly more cases...

I will try to add some tests hitting above cases.

chongxiaoc:
@selitvin I've refactored the async dataloader into a new class. Please take a look when you have time and let me know what you think about. Thanks.

petastorm/pytorch.py
```diff
@@ -423,3 +425,160 @@ def __enter__(self):
...
+            if batch is None:
+                # None is last data in async batch queue, which means that async shuffle thread
+                # has reached the end of data.
+                self._started = False
+                break
+            yield batch
+
+        self._shuffle_thread.join()
+
+    def _iter_impl_worker(self):
```
selitvin, line 508: Appears that _iter_impl_worker has exactly the same implementation as its baseclass's: `BatchedDataLoader` `_iter_impl` implementation. Can we avoid this copy/paste somehow?

petastorm/pytorch.py
```diff
@@ -423,3 +425,160 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.reader.stop()
         self.reader.join()
+
+
+class AsyncBatchedDataLoader(BatchedDataLoader):
```
selitvin, line 430: It's a matter of personal style, but I'd typically avoid functional inheritance when not really-really appropriate. The main reason is that the interface between inherited and the parent object is not easy to track. For example, it took me some time to understand which methods of the parent are being called and which methods are reimplemented in this class.

I wonder, how would the code look like, if the reusable functions would be just free functions? E.g. we could define a free function
```python
def _yield_batches(buffer, keys) :
```
that could be used by both `BatchedDataLoader` and `AsyncBatchedDataLoader` even if we drop the inheritance relation.

Would also need to make `_instantiate_buffers` and `_iter_impl_worker` free functions. 

I won't insist on this design, but curious what are your thoughts on this?

petastorm/pytorch.py
```diff
@@ -423,3 +425,160 @@ def __enter__(self):
...
+                                                     shuffling_queue_capacity=shuffling_queue_capacity,
+                                                     inmemory_cache_all=inmemory_cache_all,
+                                                     num_epochs=num_epochs)
+        self._async_batch_queue = Queue(async_batch_queue_capacity)
+        self._finished_event = Event()
+        self._shuffle_thread = None
+        self._started = False
+
+    def _worker(self):
+        for batch in self._iter_impl_worker():
```
selitvin, line 485: What happens when `self._iter_impl_worker` raises an exception? Could it be that `batch = self._async_batch_queue.get()` will be stuck forever, or there is some mechanism that prevents that?

Let's cover the exception raising in a unittest? (could put some raise in a transform func to make it easy to test)

petastorm/pytorch.py
```diff
@@ -423,3 +425,160 @@ def __enter__(self):
...
+            set to None, loader will loads batches indefinitely.
+        :param async_batch_queue_capacity: Capacity of queue to store shuffled batches. Shuffle work is done by an
+            asynchronous thread to produce batches into a queue.
+        """
+        super(AsyncBatchedDataLoader, self).__init__(reader, batch_size=batch_size, transform_fn=transform_fn,
+                                                     shuffling_queue_capacity=shuffling_queue_capacity,
+                                                     inmemory_cache_all=inmemory_cache_all,
+                                                     num_epochs=num_epochs)
+        self._async_batch_queue = Queue(async_batch_queue_capacity)
+        self._finished_event = Event()
```
selitvin, line 480: It seems the event is being used as a simple boolean flag (we don't wait on it, e.g.). Would a simple bool work here (as it is a simpler, sufficient construct)

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -248,65 +253,93 @@ def test_batched_data_loader_with_in_memory_cache(two_columns_non_petastorm_data
...
+
+            if num_epochs is None:
+                for idx in range(5):
+                    batch = next(it)
+                    this_batch = batch['col_0'].clone()
+                    assert list(this_batch.shape)[0] == batch_size
+                    retrieved_so_far = torch.cat([retrieved_so_far, this_batch], 0)
+
+
+def test_async_batched_data_loader(two_columns_non_petastorm_dataset):
```
selitvin, line 321: Can you please document the test a bit - what's the purpose of the test and what it's testing.

I know `test_batched_data_loader_with_in_memory_cache` is the test that was there before, but if you understand exactly how it works, would appreciate if you could paste that knowledge as comments in the test. I have hard time parsing it :). 

If you haven't dive deep into it, then never mind.

petastorm/pytorch.py
```diff
@@ -423,3 +425,160 @@ def __enter__(self):
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # Close async shuffle thread
+        if self._shuffle_thread:
+            self._close_async_thread()
+        super(AsyncBatchedDataLoader, self).__exit__(exc_type, exc_val, exc_tb)
+
+    def __del__(self):
+        # Close async shuffle thread
+        if self._shuffle_thread:
+            self._close_async_thread()
```
selitvin, line 572: If possible, please test-cover this code path?

chongxiaoc: This could be done in a test without using context manager.

petastorm/pytorch.py
chongxiaoc: `_iter_impl()` is from parent class that base dataloader uses to iterate, maybe I can do super. _iter_impl() for `thread` function, I will try.

petastorm/pytorch.py
chongxiaoc: This seems very neat, I will think about it.

petastorm/tests/test_pytorch_dataloader.py
chongxiaoc: sure. Maybe this will be removed after #664 is resolved.

chongxiaoc:
convert to draft for now.

chongxiaoc:
close this PR since asyncp process is added into Horovod

Issue #655, opened 2021-03-17
Remove implementation of __getattr__ for UniSchema


v01dXYZ:
Hello,
I found a very minor issue (probably a typo):
https://github.com/uber/petastorm/blob/d6f4e82eb2c3a6c2b4c16c060c7350331b60a51a/petastorm/unischema.py#L355-L356
This is not really required (basically since we don't reimplement `__getattribute__`, that will always be `object.__getattribute__(obj, item)` that will be invoked when we `getattr`).

selitvin:
Thanks for the report. Will clean it up (or feel free to submit a fix in a PR :) ).

PR #656, opened 2021-03-18
Remove Unischema __getattr__ implementation

v01dXYZ:
This MR solves #655.

It removes the implementation of the fallback `Unischema.__getattr__` which doesn't seem to add any value.


selitvin:
I see that the build is broken for some unrelated reason. Fixing it.

PR #659, opened 2021-03-19
Fixate numpy version used in build

selitvin:
Having a loose `pip install -U numpy` in .travis.yml resulted in
non-deterministic version of numpy and as a result a broken build.

Currently fixing the version to an older numpy 1.19 and will upgrade CI
versions of pyarrow and numpy in cohort in the following PRs.


chongxiaoc (approved):
LGTM

PR #662, opened 2021-04-09
Reader: shuffle row groups before sharding.

chongxiaoc:
Splitting row groups by % num_shards is not a good idea to split data.
This commit introduce a shard_seed parameter to random shuffle row
groups before sharding.

### DIFF
--- a/petastorm/reader.py
+++ b/petastorm/reader.py
@@ -65,7 +65,7 @@ def make_reader(dataset_url,
                 predicate=None,
                 rowgroup_selector=None,
                 num_epochs=1,
-                cur_shard=None, shard_count=None,
+                cur_shard=None, shard_count=None, shard_seed=None,
                 cache_type=NULL_CACHE, cache_location=None, cache_size_limit=None,
                 cache_row_size_estimate=None, cache_extra_settings=None,
                 hdfs_driver='libhdfs3',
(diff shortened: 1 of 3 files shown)


petastorm/reader.py
```diff
@@ -378,8 +378,10 @@ def __init__(self, pyarrow_filesystem, dataset_path, schema_fields=None,
         :param filters: (List[Tuple] or List[List[Tuple]]): Standard PyArrow filters.
             These will be applied when loading the parquet file with PyArrow. More information
             here: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html
+        :param shard_seed: Random seed to shuffle row groups for data sharding. Defaults to None
         """
         self.num_epochs = num_epochs
+        self.shard_seed = shard_seed
```
selitvin, line 384: Since we use this value only from the constructor , suggest we don't turn it into a state variable, but pass it down the call-stack to the consumer. By limiting the scope of the variable, it makes the code more readable, IMO.

selitvin (changes requested):
Please add a unit test to cover the new case.

petastorm/reader.py
chongxiaoc: sounds reasonable to me.

chongxiaoc:
@selitvin I addressed the comments, can you take one more look?

petastorm/reader.py
```diff
@@ -550,6 +556,13 @@ def _partition_row_groups(self, dataset, row_groups, shard_count, cur_shard,
 
         # We hash on the relative path of each parquet file to guarantee consistency between different reader
         # constructions even after moving the dataset
+        if shard_seed:
+            import random
```
selitvin, line 560: Do you know why these lines appear as not covered? Is anything wrong with either the test or the code?

chongxiaoc: hmmm, I modified one unit test to cover `shard_seed` != None case. Let me double check.

chongxiaoc: Should be fixed. I used `0` for shard_seed, which make `if shard_seed` false.

chongxiaoc:
@selitvin please take a look again, fixed codecov issues.

selitvin (approved):
Sorry, forgot to ask for one last thing. Please update the release notes (docs/release-notes.rst).

chongxiaoc:
@selitvin updated release notes

PR #663, opened 2021-04-10
fix  get_dataset_path() in fs_utils.py

dongpohezui:
version:
Windows 10
Python 3.7.0
petastorm 0.9.8
pyarrow 3.0.0

```
from petastorm import make_reader

reader = make_reader('file:///D:/test/test.parquet') 
print(reader)
```
The following error occurred when I ran the above code under Windows.

picture

The image above shows that there is an extra '/' at the beginning of the file path.
So I deleted the extra '/' .

### DIFF
--- a/petastorm/fs_utils.py
+++ b/petastorm/fs_utils.py
@@ -16,6 +16,7 @@ import logging
 import pyarrow
 import six
 from six.moves.urllib.parse import urlparse
+import platform
 
 from petastorm.gcsfs_helpers.gcsfs_wrapper import GCSFSWrapper
 from petastorm.hdfs.namenode import HdfsNamenodeResolver, HdfsConnector
@@ -32,6 +33,9 @@ def get_dataset_path(parsed_url):
     if parsed_url.scheme.lower() in ['s3', 's3a', 's3n', 'gs', 'gcs']:
         # s3/gs/gcs filesystem expects paths of the form `bucket/path`
         return parsed_url.netloc + parsed_url.path
+
+    if parsed_url.scheme.lower() in ['file'] and "Windows" in platform.system():
+        return parsed_url.path[1:]
 
     return parsed_url.path
 


selitvin:
Did you manage to track down what is the origin of this extra '/'? Which part of code added it? Asking this to make sure the same issue wouldn't pop up in some other scenarios with different code paths.

Also, can you please add a unittest covering this case?

dongpohezui:
The URL comes from user input.
Under Linux, it works just fine.
Under Windows, it makes mistakes.

In get_filesystem_and_path_or_paths(), URLparse () is used to handle URLs.

Under Windows, the file cannot be accessed via the file path "/D:/test/test.parquet ". Instead, it should be "D:/test/test.parquet".

In my opinion, you can only test this code under Windows.

petastorm/fs_utils.py
```diff
@@ -33,6 +34,9 @@ def get_dataset_path(parsed_url):
         # s3/gs/gcs filesystem expects paths of the form `bucket/path`
         return parsed_url.netloc + parsed_url.path
 
+    if parsed_url.scheme.lower() in ['file'] and "Windows" in platform.system():
+        return parsed_url.path[1:]
```
selitvin, line 38: There is a chance that this fix would break for file urls on network location (as per this [article](https://en.wikipedia.org/wiki/File_URI_scheme#Windows)). Consider the example of the following WIndows url: file://hostname/path/to/the%20file.txt

petastorm/fs_utils.py
```diff
@@ -33,6 +34,9 @@ def get_dataset_path(parsed_url):
         # s3/gs/gcs filesystem expects paths of the form `bucket/path`
         return parsed_url.netloc + parsed_url.path
 
+    if parsed_url.scheme.lower() in ['file'] and "Windows" in platform.system():
```
selitvin, line 37: It would be great if you could add a unit test verifying correctness of this function in `petastorm/tests/test_fs_utils.py`. You would need to mock the value returned by `platform.system()` and then I think the test logic should be straight forward.

If for some reason you can not do this, we'd be able to land this and I can help with the test later.

selitvin (changes requested):
Marking the PR with "Request Changes" due to incorrect behavior in the case of windows urls that contain network and not local paths.

PR #665, opened 2021-04-23
Replaced s3 and gcs connectors with fsspec to support additional filesystems

tgaddair:
[fsspec](https://filesystem-spec.readthedocs.io/en/latest/intro.html) is a library used by Pandas and Dask to enable reading and writing to arbitrary filesystems (see [here](https://github.com/intake/filesystem_spec/blob/master/fsspec/registry.py#L88) for a list of filesystems supported out of the box).

One of the key features of fsspec is its interoperability with `pyarrow` filesystems by design. As such, any filesystem created by fsspec can be used to read the `ParquetDataset` provided by pyarrow.

This PR replaces existing implementations of S3 and GCS filesystems in Petastorm by delegating to fsspec instead, adding additional support for other similar object storage filesystems like ADLS gen2 (MS Azure equivalent). It also replaces the `s3_config_kwargs` with the more general `storage_options` used by Pandas and Dask.


selitvin (commented):
Can  you please add some info into docs/release_notes.rst with porting instructions? 
With which systems did you test this code?

tgaddair:
Thanks for taking a look @selitvin! I'm testing this with the following OS:

- macOS
- Linux

And the following filesystems:

- S3
- GCS
- ADLS

I would suggest holding off on merging (if it looks good) until Monday, as I still want to run a few more tests this weekend to make sure everything is correct.

selitvin (approved):
Ok. Please ping me when you want to merge. Let me know when you want to cut a release.

tgaddair:
Hey @selitvin, thanks I verified that all of the environments worked as expected. Please feel free to merge whenever you're ready. Also, creating a new release would be awesome.

PR #669, opened 2021-04-28
Pytorch: add inmemory batched dataloader

chongxiaoc:
Including two changes:
- Remove current inmem_cache option for pytorch batched dataloader
- Create a new class for inmemory batched dataloader

Related to issue#664


petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+            n_rows = expected_rows
+
+        # At this point, dataloader has enough rows storted in memory.
+        # Set last_row_consumed = True in reader rather than draining the remainder
+        # of reader. If reader has infinite epochs, draining will be deadlock.
+        for row in self.reader:
+            self.reader.last_row_consumed = True
+        # Truncate empty tensors if capacity is larger than total rows.
+        if n_rows < self.capacity:
+            for k, v in enumerate(self.buffer):
```
selitvin, line 455: nit: Perhaps `for i, v in ...` to help reader understand that this is an iteration number?

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+            # Copy new items into buffer
+            for k, v in enumerate(items):
+                self.buffer[k][n_rows:expected_rows] = v[:last_row]
+            n_rows = expected_rows
+
+        # At this point, dataloader has enough rows storted in memory.
+        # Set last_row_consumed = True in reader rather than draining the remainder
+        # of reader. If reader has infinite epochs, draining will be deadlock.
+        for row in self.reader:
+            self.reader.last_row_consumed = True
```
selitvin, line 452: I think what you want to do instead of this `for` loop is `self.reader.stop()`.

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+
+            if self.capacity <= expected_rows:
+                buffer_full = True
+                last_row = self.capacity-n_rows
+                expected_rows = self.capacity
+            if self.buffer is None:
+                # Initialize buffer as a list of empty tensors
+                self.buffer = []
+                for v in items:
+                    self.buffer.append(torch.empty((self.capacity,) + v.shape[1:], dtype=v.dtype, device=v.device))
```
selitvin, line 442: Can you please clarify what does the `+v.shape[1:]` term purpose?

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+
+    def _load_rows_into_mem(self):
+        n_rows = 0
+        buffer_full = False
+        self.buffer = None
+        self.keys = None
+
+        for row in self.reader:
+            if buffer_full:
+                break
```
selitvin, line 412: It's easy for a user to assume that if they use this class all data going to be loaded into memory. They might not notice the `capacity` constructor argument since it has a default value. I think we should make `capacity` a mandatory argument or find some other API change that make this behavior of dropping rows more obvious. 

Perhaps tweaking the name to hint that the value is given as rows (and not bytes), e.g. rows_capacity.

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+                self.buffer[k] = self.buffer[k][:n_rows]
+
+    def _iter_impl(self):
+        """
+        The Data Loader iterator stops the for-loop when num_epochs is reached.
+        """
+        self._load_rows_into_mem()
+
+        for epoch in range(self.num_epochs):
+            self.size = len(self.buffer[0])
```
selitvin, line 465: `size` can be a local variable. No reason for it to be the class state from what I see.

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+                batch = dict(zip(self.keys, batch))
+                yield batch
+            assert self.size == 0
+
+    # Functions needed to treat data loader as a context manager
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.reader.stop()
```
selitvin, line 488: Your logic shuts the reader earlier - once all data was read. It's a good thing since we release resources we would no longer need. No need to `stop()` the reader here, but ok to join to make sure the `InMemBatchedDataLoader` indeed acts as  a context-manager and does not exit until everything is clean.

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+        self.reader = reader
+        self.batch_size = batch_size
+        self.transform_fn = transform_fn or torch.as_tensor
+        self.num_epochs = num_epochs
+        self.seed = seed
+        self.capacity = capacity
+        self.shuffle = shuffle
+        self.size = 0
+
+    def _load_rows_into_mem(self):
```
selitvin, line 404: Would making this a standalone (non class function) that returns `(keys, buffers) ` tuple more clear and easier to reuse if needed?
```python
_loader_rows_into_mem(reader, transform_fn, capacity)
```
We would need to `stop()` the reader in iter_impl though.

Having buffer and keys part of the class state is unnecessary since it is not used after `iter_impl` exists (hence it would be good to release the memory). Or am I wrong and `InMemBatchedDataLoader`  could be used multiple times somehow?

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+        # Truncate empty tensors if capacity is larger than total rows.
+        if n_rows < self.capacity:
+            for k, v in enumerate(self.buffer):
+                self.buffer[k] = self.buffer[k][:n_rows]
+
+    def _iter_impl(self):
+        """
+        The Data Loader iterator stops the for-loop when num_epochs is reached.
+        """
+        self._load_rows_into_mem()
```
selitvin, line 462: How about:
```python

if self.buffer is not None:
  buffer, keys = _load_rows_into_mem(...)
```
this way, we:
 1. don't need to reread the data as in `test_call_iter_on_dataloader_multiple_times_inmem` test case
 2. can safely stop the read after the loading of the data finished first time.

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -208,3 +256,68 @@ def test_call_iter_on_dataloader_multiple_times(synthetic_dataset, reader_factor
...
+
+@pytest.mark.parametrize('reader_factory', MINIMAL_READER_FLAVOR_FACTORIES)
+def test_inmem_batched_dataloader_shuffle_per_epoch(synthetic_dataset, reader_factory):
+    """See if we are getting different shuffled result in different epochs"""
+    epochs = 3
+    with InMemBatchedDataLoader(reader_factory(synthetic_dataset.url, schema_fields=['^id$']),
+                                batch_size=10, capacity=100, num_epochs=epochs, shuffle=True) as loader:
+        first_result = []
+        second_result = []
+        third_result = []
```
selitvin, line 301: Unused variable

petastorm/pytorch.py
```diff
@@ -367,3 +367,123 @@ def __enter__(self):
...
+        :param capacity: number of rows to be loaded into memory.
+        :param shuffle: If ``True``, indices will be shuffled in every epoch.
+        """
+        super(InMemBatchedDataLoader, self).__init__()
+        self.reader = reader
+        self.batch_size = batch_size
+        self.transform_fn = transform_fn or torch.as_tensor
+        self.num_epochs = num_epochs
+        self.seed = seed
+        self.capacity = capacity
```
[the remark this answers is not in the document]
chongxiaoc: sure thing, I was following existing style. let me put them all private.

petastorm/pytorch.py
chongxiaoc: We want to allocate empty tensor with exact dimension of input rows.
v.shape[0] is row group size, while v.shape[1:] is possible dimensions of input row data.
This is same as code in shuffling buffer:  https://github.com/uber/petastorm/blob/7f37e8dde6ff1b13f055d22a6289e2de8bb5d473/petastorm/reader_impl/pytorch_shuffling_buffer.py#L269

chongxiaoc:
TODO: mention the class in the main Readme and release notes

petastorm/pytorch.py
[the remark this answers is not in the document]
chongxiaoc: It turns out that I have to set `reader` is public attribute, since inmem loader class inherit from `LoaderBase`, which users `.reader`.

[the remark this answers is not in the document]
chongxiaoc: Well, if `InmemDataLoader` doesn't support multiple iterations, it doesn't have to inherit `LoderBase`.

chongxiaoc:
@selitvin ready to be reviewed again.

petastorm/pytorch.py
```diff
@@ -423,3 +367,139 @@ def __enter__(self):
...
+    def __init__(self, reader, batch_size=1,
+                 transform_fn=None,
+                 num_epochs=1,
+                 seed=0,
+                 rows_capacity=1024,
+                 shuffle=False):
+        """
+        Initializes a data loader object.
+        This class does not support special types that are not supported in PyTorch (decimal/string).
+        :param reader: petastorm Reader instance
```
selitvin, line 450: Perhaps mention here that the reader will be `stop`ped once all data is loaded from it.

petastorm/pytorch.py
```diff
@@ -423,3 +367,139 @@ def __enter__(self):
...
+    def __iter__(self):
+        """
+        The Data Loader iterator stops the for-loop when num_epochs is reached.
+        """
+        if self._in_iter:
+            raise RuntimeError("InMemBatchedDataLoader couldn't be used multiple times, please\
+                    specify total number of epochs using num_epochs in constructor.")
+        self._in_iter = True
+        try:
+            keys, buffer = _load_rows_into_mem(self._reader, self._transform_fn, self._rows_capacity)
```
selitvin, line 477: Consider adding a bit more comments describing structure of `keys` and `buffer`. I think it may help a reader understand the code better before digging into `_load_rows_into_mem` implementation.

petastorm/pytorch.py
```diff
@@ -423,3 +367,139 @@ def __enter__(self):
...
+                # Sample batches
+                for i in range(0, len(indices), self._batch_size):
+                    idx = indices[i:i+self._batch_size]
+                    batch = [v[idx] for v in buffer]
+                    size -= len(batch[0])
+                    batch = dict(zip(keys, batch))
+                    yield batch
+                assert size == 0
+        except Exception as e:
+            logger.error('Iteration on InMemBatchedDataLoader raise error: %s', repr(e))
```
selitvin, line 497: What is the added value of this generic exception handler?

petastorm/pytorch.py
```diff
@@ -423,3 +367,139 @@ def __enter__(self):
...
+            raise
+
+    # Functions needed to treat data loader as a context manager
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self._reader.join()
```
selitvin, line 505: Would we get into a deadlock if a user instantiated `InMemBatchedDataLoader` but never iterated? Perhaps should unit-test that case.

petastorm/tests/test_pytorch_dataloader.py
```diff
@@ -9,10 +8,12 @@
...
 
 from petastorm import make_reader, TransformSpec, make_batch_reader
-from petastorm.pytorch import _sanitize_pytorch_types, DataLoader, BatchedDataLoader, decimal_friendly_collate
+from petastorm.pytorch import (_sanitize_pytorch_types, DataLoader,
+                               BatchedDataLoader, decimal_friendly_collate,
+                               InMemBatchedDataLoader, _load_rows_into_mem)
 from petastorm.tests.test_common import TestSchema
 
-ALL_DATA_LOADERS = [DataLoader, BatchedDataLoader]
+OOM_DATA_LOADERS = [DataLoader, BatchedDataLoader]
```
selitvin, line 16: Does OOM stands for out-of-memory? Not sure what would be a better name, but it's a bit strange to see it here :)

petastorm/pytorch.py
```diff
@@ -423,3 +367,139 @@ def __enter__(self):
...
+        self._rows_capacity = rows_capacity
+        self._shuffle = shuffle
+        self._in_iter = False
+
+    def __iter__(self):
+        """
+        The Data Loader iterator stops the for-loop when num_epochs is reached.
+        """
+        if self._in_iter:
+            raise RuntimeError("InMemBatchedDataLoader couldn't be used multiple times, please\
```
selitvin, line 473: If the data loader can not be restarted, then what do you think about a possibility of moving 
```python
self._keys, self._buffer = _load_rows_into_mem(self._reader, self._transform_fn, self._rows_capacity)
```
into the constructor? Then time to the first sample might be improved 
 - for the cases when does something heavy between construction and first read
 - step-through interactive debugging.

petastorm/pytorch.py
chongxiaoc: I'm not sure if `try` block should be used TBH. I was seeing the `LoaderBase` having `try` block for `__iter__`. I feel the only exception may come from `_load_rows_into_mem ` with OOM or reader corruption? Any suggestions?

selitvin: It could also come from within reader errors. In LoaderBase, this PR: #544 added the try/except and the goal was to raise an explicit error in the second iteration. I think the goal was to support multiple `iter(data_loader)` calls, but fail the next time if previous iterations ended up with an exception.

petastorm/tests/test_pytorch_dataloader.py
chongxiaoc: Or name them as `BASIC_DATA_LOADERS`?

selitvin: That works. Just was surprised to see an OOM here :)

petastorm/pytorch.py
chongxiaoc: sure. I was thinking about this before. Then I though it would not work out with multiple iterations.
Now since we understand that multiple iterations are not allowed, then it should be moved into constructor.

petastorm/pytorch.py
chongxiaoc: Hmm, after a second thought, I decide remove `try` block since `_in_iter` is already checked and raised if possible.

petastorm/pytorch.py
chongxiaoc: - reader is loaded in constructor
- reader is stopped after expected rows are loaded
- work_pool is joined in reader.stop(), here we can leave it nothing just `pass`.
Maybe this is more straightforward?

chongxiaoc: I added a timeout test to check possible deadlock.

chongxiaoc:
Addressed new comments. PTAL one more time. @selitvin

docs/release-notes.rst
```diff
@@ -6,6 +6,7 @@ Release notes
 
 Release 0.10.1 (unreleased)
```
selitvin, line 7: Need to remember and bump to 0.11.0 since we are changing the interface of `BatchedDataLoader` (removing arguments). No action for this diff.

selitvin (approved):
Cool stuff! Ship it!

docs/release-notes.rst
chongxiaoc: sure thing.

PR #674, opened 2021-05-05
tf_dataset: add unit test to verify repeat() works after cache()

chongxiaoc:
We can cache first iteration result then use repeat() to do multiple
epochs.

### DIFF
--- a/petastorm/tests/test_tf_dataset.py
+++ b/petastorm/tests/test_tf_dataset.py
@@ -101,8 +101,8 @@ def test_with_one_shot_iterator(synthetic_dataset, reader_factory):
 @create_tf_graph
 def test_with_dataset_repeat(synthetic_dataset, reader_factory):
     """``tf.data.Dataset``'s ``repeat`` should not be used on ``make_petastorm_dataset`` due to high costs of
-    ``Reader initialization``. A user should use ``Reader`` built-in epochs support. Check that we raise an
-    error to alert of misuse."""
+    ``Reader initialization``. A user should use ``Reader`` built-in epochs support, or use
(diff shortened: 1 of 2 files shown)


petastorm/tests/test_tf_dataset.py
```diff
@@ -121,6 +121,34 @@ def test_with_dataset_repeat(synthetic_dataset, reader_factory):
                 sess.run(iterator)
 
 
+@pytest.mark.forked
+@pytest.mark.parametrize('reader_factory', ALL_READER_FLAVOR_FACTORIES)
+@create_tf_graph
+def test_with_dataset_repeat_after_cache(synthetic_dataset, reader_factory):
```
selitvin, line 127: Not for this PR, but we should probably try using modern TF interface (here `import tensorflow.compat.v1 as tf`). That way the tests would become less bloated as we would work with eager execution mode.

docs/release-notes.rst
```diff
@@ -6,6 +6,7 @@ Release notes
 
 Release 0.10.1 (unreleased)
 ===========================
+- `PR 674 <https://github.com/uber/petastorm/pull/674>`_: tf_dataset: add unit test to verify repeat() works after cache().
```
selitvin, line 9: nit: usually I do not document unit-tests changes in release doc as this is not user facing (unless there is a bug fix).

selitvin (approved):

docs/release-notes.rst
chongxiaoc: Will remove it.

petastorm/tests/test_tf_dataset.py
chongxiaoc: I agree, it is very painful to debug without eager execution mode.

PR #686, opened 2021-05-27
Allow opening parquet stores with unsupported types.

selitvin:
Previously, an error was raised when an unsupported type (e.g. nested structure) was found in the datasource. Relaxing the constraint so that fields with unknown types would be silently ignored - simply won't show up in the loaded data.


petastorm/tests/test_common.py
```diff
@@ -229,6 +229,7 @@ def expected_row(i):
             StructField('id_div_700', IntegerType(), False),
         ] + maybe_int_fixed_size_list_field +
         [
+            StructField('nested_struct', StructType([StructField('nested_int', IntegerType(), True)])),
```
selitvin, line 232: Sneak in a nested field that is considered currently unsupported and see that all tests still pass.

chongxiaoc (approved):
LGTM

PR #687, opened 2021-05-27
Fix a failure when reading data from a parquet file (and not a parquet directory)

selitvin:
Code was crashing since ParquetDataset's (pyarrow) partition_names attribute was set to None when the ParquetDataset object instance was created from a parquet file and not parquet directory.

### DIFF
--- a/petastorm/tests/test_parquet_reader.py
+++ b/petastorm/tests/test_parquet_reader.py
@@ -72,7 +72,7 @@ def test_simple_read(scalar_dataset, reader_factory):
 
 
 @pytest.mark.parametrize('reader_factory', _D)
-def test_simple_read(scalar_dataset, reader_factory):
+def test_simple_read_from_a_single_file(scalar_dataset, reader_factory):
     """See if we can read data when a single parquet file is specified instead of a parquet directory"""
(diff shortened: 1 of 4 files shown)


petastorm/tests/test_parquet_reader.py
```diff
@@ -70,6 +71,18 @@ def test_simple_read(scalar_dataset, reader_factory):
         _check_simple_reader(reader, scalar_dataset.data)
 
 
+@pytest.mark.parametrize('reader_factory', _D)
+def test_simple_read(scalar_dataset, reader_factory):
+    """See if we can read data when a single parquet file is specified instead of a parquet directory"""
+    assert scalar_dataset.url.startswith('file://')
```
selitvin, line 77: Although our fix affects only make_reader(..) readers, let's check that the same issue does not occur with make_batch_reader.

chongxiaoc: Unit test fails due to using same function name as line 68

chongxiaoc (changes requested):
LGTM, just fix unit test name issue.

chongxiaoc (approved):
LGTM

PR #698, opened 2021-07-05
Use pyarrow.fs.LocalFileSystem as per arrow 4.x

JayjeetAtGithub:


selitvin:
Can you please update docs/release-notes and mention this change? 
How did you test the change? Did you try using a dataset on s3/hdfs?

PR #708, opened 2021-07-28
Do not attempt to override pyarrow's dataset.partitions attribute

### LINKED ISSUE #707: pyarrow v.5.0.0 breaks Petastorm Reader
`pyarrow` just released their latest version today, which breaks Petastorm. The code that breaks is in `Reader.__init__()` in `reader.py`:

```
        self.dataset = pq.ParquetDataset(dataset_path, filesystem=pyarrow_filesystem,
                                         validate_schema=False, metadata_nthreads=10,
                                         filters=filters)
```

The error message that Petastorm gives is `Can't set attribute`.

selitvin:
Removing a hack where we manually override pyarrow's `dataset.partitions` property because of an assumption of other parts of code that the attribute is always set and iterable.

pyarrow 5.0 breaks as it protects the attribute against modification.

We fix the issue by explicitly checking for `dataset.partitions` being `None` in relevant parts of petastorm code.

Add a unittest configuration to cover pyarrow 3, 4, and 5.

Resolves #707

### DIFF
--- a/.github/workflows/unittest.yml
+++ b/.github/workflows/unittest.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        config: [pyspark-2.4, tf-1.15, pyarrow-2.0, latest,  pyarrow-0.17.1, py39]
+        config: [pyspark-2.4, tf-1.15, pyarrow-2.0, pyarrow-4.0, latest,  pyarrow-0.17.1, py39]
         include:
         - config: pyspark-2.4
           PYARROW_VERSION: "2.0.0"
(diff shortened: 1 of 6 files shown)


.github/workflows/unittest.yml
```diff
@@ -42,15 +42,22 @@ jobs:
           PYSPARK_VERSION: "3.0.0"
           ARROW_PRE_0_15_IPC_FORMAT: 0
           PY: "3.7"
-        - config: pyarrow-2.0
-          PYARROW_VERSION: "2.0.0"
+        - config: pyarrow-3.0
```
chongxiaoc, line 45: After this is changed, remember to change required check in `branch security` of github settings.
Otherwise the PR checks cannot point to the right Actions checks.
Reference: https://docs.github.com/en/github/administering-a-repository/defining-the-mergeability-of-pull-requests/managing-a-branch-protection-rule

chongxiaoc (approved):

.github/workflows/unittest.yml
chongxiaoc: @selitvin ^

selitvin: Interesting. Was not aware of this. Thanks!

PR #715, opened 2021-08-17
Fix package versions in Dockerfile

chongxiaoc:
Tensorflow2 conflicts with numpy 1.20.1

Error example is:

```
#30 26.12 ERROR: Cannot install numpy==1.20.1, pyarrow==3.0.0 and tensorflow==2.5.0 because these package versions have conflicting dependencies.
#30 26.12
#30 26.12 The conflict is caused by:
#30 26.12     The user requested numpy==1.20.1
#30 26.12     pyarrow 3.0.0 depends on numpy>=1.16.6
#30 26.12     tensorflow 2.5.0 depends on numpy~=1.19.2
#30 26.12
#30 26.12 To fix this you could try to:
#30 26.12 1. loosen the range of package versions you've specified
#30 26.12 2. remove package versions to allow pip attempt to solve the dependency conflict
```


selitvin (approved):
:+1:

PR #732, opened 2022-01-07
Get rid of RuntimeWarning when using process pool

### LINKED ISSUE #585: RuntimeWarning when using pure Python reader with process workers
I noticed the following warnings while running benchmarks with pure Pyton reader using process workers (code snippet available in https://github.com/uber/petastorm/issues/584):

```
/home/dfilipsk/miniconda3/envs/petastorm/lib/python3.6/runpy.py:125: RuntimeWarning: 'petastorm.workers_pool.exec_in_new_process' found in sys.modules after import of package 'petastorm.workers_pool', but prior to execution of 'petastorm.workers_pool.exec_in_new_process'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
```

This warning is repeated as many times as the amount of workers in the pool.

selitvin:
Restructure process_pool implementation code in a way that resolves `RuntimeWarning: 'petastorm.workers_pool.exec_in_new_process' found in sys.modules after import of package 'petastorm.workers_pool', but prior to execution of 'petastorm.workers_pool.exec_in_new_process'; this may result in unpredictable behaviou when using process pool` warning.

In this diff we break main entry point from `exec_in_new_process.py` file into two a separate `exec_in_new_process_entrypoint.py`.

Resolves #585

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -7,7 +7,7 @@ Release notes
 Release 0.11.4 (unreleased)
 ===========================
 - `PR 732 <https://github.com/uber/petastorm/pull/732>`_ (resolves `PR 585 <https://github.com/uber/petastorm/issues/585>`_): Restructure process_pool implementation code in a way that resolves ``RuntimeWarning: 'petastorm.workers_pool.exec_in_new_process' found in sys.modules after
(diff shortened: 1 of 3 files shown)


chongxiaoc (approved):
Great, I don't even know this trick.

PR #737, opened 2022-02-11
Use highest available pickle protocol when serializing

rbetz:
[Pickle protocol 5](https://www.python.org/dev/peps/pep-0574/) allows for zero-copy pickling of numpy arrays, but protocol version 4 remains the default for Python 3.8 onwards. When loading datasets with large numpy arrays and `reader_pool_type="process"`, using protocol 5 significantly improves serialization time and overall performance.

Protocol 5 is only available in Python 3.8 and onwards, so for backwards and forwards compatibility with future pickle protocol improvements we specify `pickle.HIGHEST_PROTOCOL`.

## Small test case
```
import time
import numpy as np
from petastorm.reader_impl.pickle_serializer import PickleSerializer

s = PickleSerializer()
data = [np.random.random((1000, 1000)) for _ in range(10)]
t = time.time()
for x in range(100):
  _ = s.deserialize(s.serialize(data))
print(time.time() - t)
```
master:
11.40 sec

This PR:
8.78 sec


selitvin:
The change looks good to me. Can you please update docs/release-notes.rst and accept the CLA? Once we land it, I can cut a new release.

selitvin (approved):

rbetz:
bump... Is there something I need to do to get the unit tests to run?

selitvin:
> bump... Is there something I need to do to get the unit tests to run?

Sorry. Did not notice you were blocked on the tests. I had to approve the run for a user's first time. Should be ok going forward. Also, note there is a flaky test (some test that has to do with reading from a partitioned dataset). Feel free to rerunning the build if you bump into it.

rbetz:
@selitvin I think I need permission to run the tests again. They did all pass, except for the ones you mentioned were flaky.

selitvin:
That's weird.  I thought it I am expected to approve running tests only for the first time. Will see if I can reconfigure this.

rbetz:
Thanks for your help. Looks like the tests passed this time, so it's ready to merge :)

selitvin:
Beautiful! Thank you for the PR.

Issue #740, opened 2022-03-14
On BatchedDataLoader performance


jarandaf:
Hi all,

I am trying to train a PyTorch model with a pretty big dataset (of the order of millions of samples, ~100 columns, including scalars and arrays) stored as Parquet files. After reading the docs it seems `BatchedDataLoader` should be the choice.

I have been having a look at the `BatchedDataLoader` class and despite reading parquet files in parallel with PyArrow, batches seem to be built on demand in an iterative way. This somehow does not leverage the processing power of GPUs and during training I don't observe GPU usage >20% and the GPU usage is very unstable. 

I am afraid the GPU is idle waiting all those batches to be built. Would it be possible to build them in advance?

selitvin:
Batch building implementation in BatchedDataLoader should be fairly efficient. Are you sure the slowness comes from `BatchedDataLoader`? Could it be that the data is not supplied fast enough to the `BatchedDataLoader`? Did you try tweaking parameters you pass to `make_batch_reader` (specifically `reader_pool_type` and `workers_count`)?

selitvin:
Got it. Interesting. Indeed, multiplicity of columns is tricky since it is handled by these two [loops](https://github.com/uber/petastorm/blob/master/petastorm/reader_impl/pytorch_shuffling_buffer.py#L102), it might end up pretty slow.

Couple of ideas:
1. Add a pipelining - a thread + a queue: this way batch construction would be done in parallel with training. I think it's not a large undertaking and can be done externally to petastorm; alternatively, we can think of adding this as a built-in feature into `BatchedDataLoader`.
2. Your proposal of moving the batching into worker processes/threads might be doable. I am afraid it would be a bit trickier to implement. Also, a concern of having shuffling done at a per-worker basis would result in worse shuffling quality.

If you are interested, feel free to propose a PR - we can work together to get it into the petastorm codebase.

jarandaf:
Could you please elaborate on 1?

selitvin:
In (1) I am referring to the following idea:
- Implement a class that has the same interface as `BatchedDataLoader`
- The is instantiated with an instance of a petastorm's `BatchedDataLoader`
- In the constructor it instantiates a queue (bounded size) and a thread
- On the thread, we continuously read batches from `BatchedDataLoader` and store results in the queue
- `__iter__` of the new class returns data from the queue.

This way, the batching will be done on the background thread. This way the main thread can drive GPU based training while CPU/GIL would be busy creating batches.

[...]

PR #748, opened 2022-03-23
Enable batch fetching in advance

jarandaf:


petastorm/pytorch.py
```diff
@@ -118,7 +147,10 @@ def __iter__(self):
         self._in_iter = True
 
         try:
-            for batch in self._iter_impl():
+            iterator = self._iter_impl()
+            if self._max_prefetch > 1:
+                iterator = BackgroundIterator(iterator, prefetch=self.max_prefetch)
```
selitvin, line 152: Will the `BackgroundIterator` be properly destroyed and the thread joined when we exit the function (either nominal exit or with an exception?).

petastorm/pytorch.py
```diff
@@ -100,11 +102,38 @@ def decimal_friendly_collate(batch):
 loader."
 
 
+class BackgroundIterator(threading.Thread):
+    """Prefetch iterator results. A thread iterates the original iterator and
+    populates a queue. Iterating over this background iterator just consumes the underlying
+    queue until no other result is available."""
+    def __init__(self, iterator, prefetch=1000):
```
selitvin, line 109: Is it really prefetching? Setting queue size does not guarantee that we will prefetch the data before user starts consuming it, does it? Perhaps we call the argument 'queue_size'? 
Frankly, I am not sure how prefetching helps steady-state throughput. Wouldn't it just eliminate some hiccups when the training starts at the expense of training starting a bit later? Isn't steady state throughput the only important characteristic here?

selitvin, line 109: Having default value of 1000 batches for the queue size maybe a bit too much, given a batch is a row-group, and a rowgroup of couple of hundreds MBs are common.

selitvin (commented):
Looks like we are in the right direction! Left some questions.

petastorm/pytorch.py
jarandaf: Loader's `_iter_impl` yields batches and not row groups, right? This is what is enqueued. It is true that depending on the queue size more or less row groups will be processed, but I expect this to be controlled via the queue size and the batch size.

jarandaf: You are right, probably prefetching is not the right word. As discussed in #740, the main motivation of this PR is to enable parallel batch building while training a model (otherwise the model will always have to wait for a batch to be available and this may take some time, specially if the dataset has a big number of columns). I have observed a ~3x speedup in data throughput with this change.

petastorm/pytorch.py
jarandaf: Need to check this, thank you for the heads up.

petastorm/pytorch.py
selitvin: Sure - I see how this can speed up the training. This is a good change. 

In my understanding we are not really doing prefetching here (depending on the timing, the consumer might try to fetch the first batch before the thread has populated it, i.e. nothing was prefetched). 

If you are ok with just changing the name from prefetching to queue size, everything will fall in place then.

selitvin: You are absolutely right.

petastorm/pytorch.py
selitvin: I know it could be tricky, but would be good to also test these aspects in a unit test or two.

jarandaf:
@selitvin could you please have a look? Thank you!

petastorm/pytorch.py
```diff
@@ -100,11 +102,42 @@ def decimal_friendly_collate(batch):
...
+        self.name = "background_iterator"
+        self.queue = Queue(queue_size)
+        self.iterator = iterator
+        self.stop = threading.Event()
+        self.start()
+
+    def run(self):
+        while not self.stop.isSet():
+            for item in self.iterator:
+                self.queue.put(item)
```
selitvin, line 120: I don't think we can use blocking puts and gets and end up with a solution that is robust to deadlocks. Let's see if this works:
- Producer ended up filling the queue and waits on a blocking put.
- Consumer fails. Calls `iterator.stop.set()`, however since .put is blocking within an iterator queue, the event is never checked and the thread is not shut down.

Another scenario: 
- The queue is empty, hence consumer waits on a blocking `.get`. 
- However, producer raises an exception. The thread dies and the consumer is stuck forever on a .get.

I think a robust implementation for a `BackgroundIterator` could get pretty tricky. All these edge cases need to be carefully tested as these kind of failures would be hard to catch in production.

petastorm/pytorch.py
```diff
@@ -117,15 +150,20 @@ def __iter__(self):
...
+            for batch in iterator:
                 yield batch
         except Exception as e:
             self._error = e
             logger.error('Iteration on Petastorm DataLoader raise error: %s', repr(e))
             raise
         finally:
             self._in_iter = False
+            if isinstance(iterator, BackgroundIterator):
+                iterator.stop.set()
```
selitvin, line 166: Let's make `stop` a private member (i.e. `_stop`) and add an API to the BackgroundIterator that performs stop (encapsulation principle).

petastorm/pytorch.py
```diff
@@ -100,11 +102,42 @@ def decimal_friendly_collate(batch):
...
 
 
+class BackgroundIterator(threading.Thread):
+    """Prefetch iterator results. A thread iterates the original iterator and
+    populates a queue. Iterating over this background iterator just consumes the underlying
+    queue until no other result is available."""
+    def __init__(self, iterator, queue_size=1000):
+        threading.Thread.__init__(self)
+        self.name = "background_iterator"
+        self.queue = Queue(queue_size)
```
selitvin, line 112: Let's mark all data members that are not intended to be exposed to BackgroundIterator users as private (`_` prefix).

selitvin (changes requested):

chongxiaoc:
Just noticed this nice work. @jarandaf Thanks!
Is it doable to support shuffle to be parallel as well?
I think usually shuffling is the bottleneck and petastorm uses single thread for that.

PR #751, opened 2022-04-09
Upgrade CI to use latest packages of tf,pyarrow,numpy in 'latest' CI configuration

selitvin:
Also dropped archaic pyarrow 0.17.1 from the CI.


chongxiaoc (approved):

chongxiaoc:
It looks like once this PR is merged, we need to delete `unittest (pyarrow-0.17.1)` and `unittest (py39)` in settings, otherwise they will keep showing pending there.

Issue #613, opened 2020-10-19
Make use of the new pyarrow.dataset functionality instead of ParquetDataset


jorisvandenbossche:
In the Apache Arrow project, we have been working the last year on a new Dataset API. (original design document: https://docs.google.com/document/d/1bVhzifD38qDypnSjtf8exvpP3sSB5x_Kw9m-n66FB2c/edit, current python docs: https://arrow.apache.org/docs/python/dataset.html). 

The dataset API has the goal to handle (scan/materialize) data sources larger than memory, and specifically provide:

- a unified interface for different sources (different file systems (local, cloud), different file formats (Parquet, CSV, JSON, feather, ..), and potentially also other sources like odbc connection, but this is not yet implemented)
- discovery of sources (crawling directories, handle directory-based partitioned datasets, basic schema normalization, ..)
- optimized reading with pedicate pushdown (filtering rows), projection (selecting columns), parallel reading or ability to manage scan tasks yourself (eg with dask)

Think: the current pyarrow.parquet.ParquetDataset functionality, but then not specific to parquet (currently also Feather ands CSV are supported), not tied to python (so for example the R bindings of arrow also use this), and with more features (better schema normalization, more partitioning schemes, predicate pushdown / row group filtering, etc).

A small illustrative example of how it can look like in python using the new generic API:

```python
import pyarrow.dataset as ds
dataset = ds.dataset("path/to/partitioned/dataset",
                     format="parquet", partitioning="hive", filesystem=..)
table = dataset.to_table(columns=['col1', 'col2'],
                         filter=(ds.field('key2') == 1) & (ds.field('key2'== 2))
```

In addition, we also support using it from the existing pyarrow.parquet API with a keyword:

```python
import pyarrow.parquet as pq
pq.ParquetDataset("path/to/partitioned/dataset", use_legacy_dataset=False)
```

This `ParquetDataset` "shim", however, does not support the full existing API. So for example, it does not support the `.pieces` and ParquetDatasetPiece etc, which are specifically APIs that dask is using.

Long term, we would like to have this new Datasets implementation replace the python implementation of `pyarrow.parquet.ParquetDataset` (the basic `pyarrow.parquet.read/write_table` and `pyarrow.parquet.ParquetFile` are normally there to stay, but the Python ParquetDataset implementation hopefully not, long term, as it is duplicating the new Datasets implementation). 

[...]

PR #754, opened 2022-04-28
Wrapper for Arrow Datasets & Dataset Pieces

aperiodic:
This is a wrapper around PyArrow's ParquetDataset and ParquetDatasetPieces, the first part of the effort to support PyArrow's new Dataset API that was [discussed in issue #613](https://github.com/uber/petastorm/issues/613#issuecomment-1086425134).

Since this wrapper has no functionality of its own, I didn't add unit tests specifically for the wrapper; let me know if I should.

I was unable to get the tests in `petastorm/tests/test_tf_dataset.py` to complete when running them inside the provided docker container, even when I left them to run for over 24 hours. Given that this is only a wrapper, and the code paths where the wrapper is used are covered by other tests, I think it's unlikely that tests in that file would fail when all other tests pass, but I'll keep an eye on the CI results. If they fail, then don't bother reviewing this until I fix them up (which may take several rounds of pushing new commits, as I can only run those tests in CI and not locally).

### DIFF
--- a/petastorm/pyarrow_helpers/dataset_wrapper.py
+++ b/petastorm/pyarrow_helpers/dataset_wrapper.py
@@ -1,0 +1,98 @@
+#  Copyright (c) 2022 BlackBerry Limited.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pyarrow import parquet as pq
+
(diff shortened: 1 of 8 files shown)


petastorm/pyarrow_helpers/dataset_wrapper.py
```diff
@@ -0,0 +1,98 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pyarrow import parquet as pq
+
+
+class PetastormPyArrowDataset:
```
selitvin, line 18: Please add a comment  explaining why do we need the wrapper.

selitvin (commented):
LGTM! Do you think we can keep it in a PR for now until we have a followup that builds on this PR with the actual new version of a dataset being used?

petastorm/pyarrow_helpers/dataset_wrapper.py
westfly: good feature

PR #757, opened 2022-06-24
Replace process_iter by pid_exists

MostafaFarahani:
`process_iter` is very slow for checking if a process is alive. `pid_exists` is more efficient.

See: https://psutil.readthedocs.io/en/latest/index.html?highlight=process_iter#psutil.pid_exists


selitvin (approved):
Very neat! Thanks for the contribution.

MostafaFarahani:
@selitvin - Thanks for the quick review!

PR #767, opened 2022-08-09
Reader: enable shuffling inside every row group

chongxiaoc:
- Allow shuffle inside every single row group in reader.
- Also use a shuffle seed to generate reproducible shuffle results across: single row shuffle, row groups shuffle in ventilator and sharding.

### DIFF
--- a/docs/release-notes.rst
+++ b/docs/release-notes.rst
@@ -7,6 +7,23 @@ Release notes
 
 Release 0.11.6 (unreleased)
 ===========================
+- `PR 767 <https://github.com/uber/petastorm/pull/767>`_: Reader: enable shuffling inside every row group.
+
+New features
+--------------------------
+- Parallel shuffling is implemented in reader level to increase throughput.
+
+  - Introduce ``shuffle_rows`` boolean field in ``Reader``, ``make_reader()`` and ``make_batch_reader()``
+    to enable shuffling inside a single row group.
+
+  - Break the bottleneck of shuffling samples in upper level PyTorch or Tensorflow dataloaders.
+
+- Support reproducible randomized shuffling outputs with ``seed`` field in ``Reader``, ``make_reader()`` and ``make_batch_reader()``.
+
+Deprecated features
+--------------------------
+- ``shard_seed`` field is deprecated in ``Reader``, ``make_reader()`` and ``make_batch_reader()``,
+  use ``seed`` field to apply randomization effects on sharding row groups.
 
 
 Release 0.11.5
--- a/petastorm/arrow_reader_worker.py
+++ b/petastorm/arrow_reader_worker.py
@@ -100,6 +100,8 @@ class ArrowReaderWorker(WorkerBase):
         self._transform_spec = args[6]
         self._transformed_schema = args[7]
         self._arrow_filters = args[8]
+        self._shuffle_rows = args[9]
+        self._random_state = np.random.RandomState(seed=args[10])
 
         if self._ngram:
             raise NotImplementedError('ngrams are not supported by ArrowReaderWorker')
@@ -287,6 +289,9 @@ class ArrowReaderWorker(WorkerBase):
 
         # pyarrow would fail if we request a column names that the dataset is partitioned by
         table = piece.read(columns=column_names - partition_names, partitions=self._dataset.partitions)
+        if self._shuffle_rows:
+            indices = self._random_state.permutation(table.num_rows)
+            table = table.take(indices)
 
         # Drop columns we did not explicitly request. This may happen when a table is partitioned. Besides columns
         # requested, pyarrow will also return partition values. Having these unexpected fields will break some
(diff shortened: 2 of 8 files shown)


chongxiaoc:
Attach some benchmark result using synthetic dataset with PyTorch:
- Experiment setup: `BatchedDataLoader + make_batch_reader()`, batch_size=50000, shuffle_buffer_size=1000000, thread_pool, 10 workers.
- Datasets from 100M rows to 1.6B rows are tested. 
- Compare throughput of shuffle in dataloader and shuffle in reader .

Synthetic Dataset PyTorch Throughput

Using shuffle in reader is expected to generate higher throughput since multiple workers are shuffling in parallel.

fyi @selitvin

petastorm/arrow_reader_worker.py
```diff
@@ -100,13 +100,17 @@ def __init__(self, worker_id, publish_func, args):
...
+        self._shuffle_seed = args[10]
 
         if self._ngram:
             raise NotImplementedError('ngrams are not supported by ArrowReaderWorker')
 
         # We create datasets lazily in the first invocation of 'def process'. This speeds up startup time since
         # all Worker constructors are serialized
         self._dataset = None
+        if self._shuffle_seed:
+            np.random.seed(seed=self._shuffle_seed)
```
selitvin, line 113: This uses global state, right? This might create conflict with other users of the random generator, running on this or other threads.If we want to setup seed, I suggest we create a seeded random number generator and use it.
```python
random_state = np.random.RandomState(42)
random_state.permutations(10)
```

petastorm/reader.py
```diff
@@ -201,11 +204,12 @@ def make_reader(dataset_url,
 def make_batch_reader(dataset_url_or_urls,
                       schema_fields=None,
                       reader_pool_type='thread', workers_count=10,
+                      shuffle_rows=False, shuffle_seed=None,
                       shuffle_row_groups=True, shuffle_row_drop_partitions=1,
                       predicate=None,
                       rowgroup_selector=None,
                       num_epochs=1,
-                      cur_shard=None, shard_count=None, shard_seed=None,
```
selitvin, line 208: Can we keep both seeds for now for the next couple of versions and add a deprecation warning? That way we don't break compatibility but give users a warning.

petastorm/reader.py
```diff
@@ -512,7 +521,7 @@ def _filter_row_groups(self, dataset, row_groups, predicate, rowgroup_selector,
         :param cur_shard: An int denoting the current shard number used. Each node should
                        pass in a unique partition number in the range [0, shard_count).
         :param shard_count An int denoting the number of reader shards
-        :param shard_seed: If not None: random seed to shuffle row groups for data sharding.
+        :param shuffle_seed: If not None: random seed to shuffle row groups for data sharding.
```
selitvin, line 524: Should we just call this `seed` since it used for multiple purposes?

petastorm/tests/test_parquet_reader.py
```diff
@@ -220,3 +220,41 @@ def test_string_partition(reader_factory, tmpdir, partition_by):
         row_ids_batched = [row.id for row in reader]
     actual_row_ids = list(itertools.chain(*row_ids_batched))
     assert len(data) == len(actual_row_ids)
+
+
+def test_shuffle_per_row_group(scalar_dataset):
+    """Check if every row group is shuffled."""
+    for reader_factory in [make_batch_reader, make_reader]:
+        with reader_factory(scalar_dataset.url, reader_pool_type='dummy',
+                            shuffle_rows=True, shuffle_row_groups=False) as reader:
```
selitvin, line 229: Do you think we need to test that shuffle_rows=False work? My guess it's covered somewhere else, just wanted to make sure. Say we hardcode it to =True in make_reader/make_batch_reader: would other tests start failing?

selitvin (commented):
Looks good! Just few minor comments. Also, please update release notes.

petastorm/arrow_reader_worker.py
chongxiaoc: Let me test. I was not sure if process pool needs to recreate a new random state.

petastorm/tests/test_parquet_reader.py
chongxiaoc: I think `shuffle_rows=False` is default case, which is tested in other existing reader tests. Those tests compare output order and length.

petastorm/arrow_reader_worker.py
chongxiaoc: addressed.

docs/release-notes.rst
```diff
@@ -7,7 +7,7 @@ Release notes
 
 Release 0.11.6 (unreleased)
 ===========================
-
+- `PR 767 <https://github.com/uber/petastorm/pull/767>`_: Reader: enable shuffling inside every row group.
```
selitvin, line 10: Can you please add just a bit more info? What's the benefit and which arg to use to enable it. Also, please mention deprecated argument.

selitvin (approved):

docs/release-notes.rst
chongxiaoc: addressed

PR #792, opened 2023-03-16
Add missing field_name in ValueError

chasleslr:
I encountered this error a few times and noticed the variable name was missing. Without this commit, the error currently reads:
```
Field {} is not "nullable", but got passes a None value
```


selitvin:
Thanks for the fix! So we can land your change, can you please accept the CLA?

PR #809, opened 2025-07-28
Update unittest.yml to run tests only on the latest versions

arushi297:
This PR does the following:
1. Update the setup tools to use version <70 to fix the CI checks
2. Remove the older pyarrow version from the checks and just keep the latest version (6.0.1)
3. Remove the codecov to unblock on the rate limit issue. We added the CODECOV_TOKEN but while it shows the check passed, it still has error uploading the code cov report and blocks the other checks to run past this


diyu2025:
looks like the PR can not pass current test rules.  After it lands it will be able to pass test rules.  So we need to bypass the rules for now.

PR #814, opened 2025-08-15
Multiple incremental improvements to local disk cache

kashishmittal55:
Multiple incremental improvements to local disk cache to prepare for Michelangelo + Ray + disk cache changes:

- Close threads in teardown before removing underlying files or closing cache.
- Use cache volume and size to detect whether to insert data in cache or not if eviction policy is none
- Eviction policy can be set to none by the client. The reason for this is, if we warm the cache in first epoch for ~40% of data, we can just keep it cached for subsequent epoch as evicting it with newer data would basically reduce our cache hit to 0

The training time improved from 34 hours to 9 hours if we use eviction policy none when data > cache_size
image
image

### DIFF
--- a/petastorm/local_disk_cache.py
+++ b/petastorm/local_disk_cache.py
@@ -37,10 +37,6 @@ class LocalDiskCache(CacheBase):
         :param settings: these parameters passed-through to the diskcache.Cache class.
                          For details, see: http://www.grantjenks.com/docs/diskcache/tutorial.html#settings
         """
-        if size_limit_bytes / shards < 5 * expected_row_size_bytes:
-            raise ValueError('Condition \'size_limit_bytes / shards < 5 * expected_row_size_bytes\' needs to hold, '
-                             'otherwise, newly added cached values might end up being immediately evicted.')
-
         default_settings = {
             'size_limit': size_limit_bytes,
             'eviction_policy': 'least-recently-stored',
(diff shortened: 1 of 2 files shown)


kashishmittal55:
@[diyu2025](https://github.com/diyu2025) diyu2025 [yesterday](https://github.com/uber/petastorm/pull/813#discussion_r2274270815)
This change also changed the cache behavior. It becomes possible that the value will be None at the final return. Do we handle such new case? This change will make cache always return None, if a new item is not in cache and the cache is full.

Author
@[kashishmittal55](https://github.com/kashishmittal55) kashishmittal55 [yesterday](https://github.com/uber/petastorm/pull/813#discussion_r2275319346)
I am sorry Di, I am confused.
So if the value from cache is None, we fill the cache value using fill_cache_func, so we will return the actual value if value is None, the behavior which is changing is, we check the cache size before puting the value in cache as eviction policy None will result cache into evergrowing and never evict.
Since we reverted the previous default policy change to None, I modified the code to only do these checks when eviction policy is None as for other cases, we will evict and store the new value.
PTAL and thanks!

kashishmittal55:
[arushi297](https://github.com/arushi297) left a comment
@kashishmittal55 do you also have performance comparison data for eviction_policy='least-recently-stored' and eviction_policy='none' in terms of cache hit rate/training speed? Would be good to have a comparison of these 2 and insights on when to use which one.


We see an improvement from 34 hours to <9 hours of training time when we move the eviction policy to none. Added more details in the PR

